PR #1, opened 2022-03-14
[HOPSWORKS-3033] Python API for projects, datasets, jobs, executions and git

robzor92:


robzor92:
Fixes #2

python/hopsworks/engine/git_engine.py
```diff
@@ -0,0 +1,62 @@
...
+            time.sleep(5)
+
+        if git_op.success is False:
+            raise GitException(
+                "Git command failed with the message: {}".format(
+                    git_op.command_result_message
+                )
+            )
+        else:
+            self._log.info("Git command {} finished".format(command))
```
gibchikafa, line 60: Here you can also log the `git_op.command_result_message`

robzor92: The API returns a List of GitFileStatus objects with the changes actually.

robzor92: For example [GitFileStatus('model_design_doc.md', 'D', '')]

PR #114, opened 2022-10-28
[HWORKS-218] Fix upload for clustered payara

ErmiasG:


ErmiasG:
hopsworks-api-upload

ErmiasG:
https://github.com/logicalclocks/hops-testing-ee/pull/578

ErmiasG:
Reverting stream upload because it moves the problem from disk to memory

PR #120, opened 2022-12-06
[HWORKS-289] Read only repositories

robzor92:


gibchikafa (approved):

gibchikafa:
Can we use hworks jira: HWORKS-289 instead of front-end?

PR #130, opened 2023-01-23
[HWORKS-381] hopsworks-api connection.close() should clean up hsfs and hsml connections

robzor92:


python/hopsworks/__init__.py
```diff
@@ -241,14 +241,6 @@ def _prompt_project(valid_connection, project):
 
 
 def logout():
-    global _connected_project
-    if _connected_project is not None:
```
moritzmeister, line 245: Would it hurt to close it here too?

robzor92: It will actually close it when doing _saas_connection.close() on the lines below, this is the connection formed when doing hopsworks.login(). This code was written under the assumption that hopsworks.login() should only be used for app hence the naming. I'll rename _saas_connection to something else.

robzor92: Pushed the rename

moritzmeister (approved):

PR #158, opened 2023-07-17
[HWORKS-351] Docker custom commands

gibchikafa:


python/hopsworks/command.py
```diff
@@ -24,6 +24,7 @@ def __init__(
         op=None,
         install_type=None,
         error_message=None,
+        custom_commands_file=None,
```
kennethmhc, line 27: `id` and `args` need to be included as well.

python/hopsworks/command.py
```diff
@@ -20,19 +20,25 @@
...
         href=None,
         count=None,
     ):
+        self._id = id
         self._op = op
         self._install_type = install_type
         self._status = status
         self._error_message = error_message
         self._count = count
+        self._custom_commands_file = custom_commands_file
```
kennethmhc, line 40: Would be nice to leave a comment why they should be added so that we don't regress.

kennethmhc (changes requested):

kennethmhc (approved):

PR #161, opened 2023-08-25
[HWORKS-695] Add blocking flag when deleting git repository

gibchikafa:


python/hopsworks/core/git_api.py
```diff
@@ -224,6 +225,34 @@ def get_repo(self, name: str, path: str = None):
         else:
             raise GitException("No git repository found matching name {}".format(name))
 
+    def _delete_repo(self, repo_id, name, verify_delete=True, timeout=30):
```
o-alex, line 228: any reason we call it `verify_delete` instead of the more familiar `blocking` or `sync`/`synchronous` or the opposite `asynchronous`/`async`

o-alex (changes requested):

python/hopsworks/core/git_api.py
o-alex: How is the delete synchronous if we have to wait and check if the directory is still there? That is the definition of asynch not sync.

gibchikafa: That's why I called the variable `verify_delete` because after the API call I expected everything to be completed. But just to make sure we can still verify.

o-alex: Wait now I am really confused... is the backend synchronous or asynchronous. Because if the backend is supposed to be synchronous, then this PR is pointless and is just covering a bug in the backend. We should fix the backend.
There is no, "we expect the dir to be gone, but just to make sure we verify" - then this is a bug...

gibchikafa: The backend is synchronous. There is no bug in the backend. The verify delete is not necessary I agree. It only makes sense if we were using the dataset API to delete the repository. But in this case I am using git api which is synchronous. I can remove the `verify_delete` stuff.

gibchikafa:
Moved to https://github.com/logicalclocks/hopsworks-api/pull/163

PR #176, opened 2023-10-15
[HWORKS-791] Add Python API support to schedule jobs

SirOibaf:


python/hopsworks/job.py
```diff
@@ -47,6 +50,11 @@ def __init__(
         self._creator = creator
         self._executions = executions
         self._project_id = project_id
+        self._job_schedule = (
+            js.JobSchedule.from_response_json(job_schedule)
+            if job_schedule
```
robzor92, line 55:
```suggestion
            if job_schedule is not None
```

python/hopsworks/job.py
```diff
@@ -180,6 +193,49 @@ def delete(self):
         """
         self._job_api._delete(self)
 
+    def schedule(self, cron_expression, start_time=None, end_time=None):
```
robzor92, line 196: To be consistent with how we show what datatypes in the docs for client facing APIs - Can you remove the `datetime, optional` etc from the docstring below and instead add the type in the function signature? I.e like this https://github.com/logicalclocks/hopsworks-api/blob/14f46bca54f4f8f5eedd5c32d5cd3b86be590b17/python/hopsworks/job.py#L120

The resulting docs will show the type for each function parameter: https://docs.hopsworks.ai/hopsworks-api/3.4/generated/api/jobs/#run

robzor92 (changes requested):

python/hopsworks/job.py
SirOibaf: In Python `if job_schedule` will return true if the object is not None or empty string.

python/hopsworks/job.py
SirOibaf: Fixed

robzor92 (approved):

PR #183, opened 2023-11-09
[HWORKS-559] Specify directory to download logs

gibchikafa:


python/hopsworks/engine/execution_engine.py
```diff
@@ -20,25 +20,33 @@
...
 class ExecutionEngine:
     def __init__(self, project_id=None):
         self._dataset_api = dataset_api.DatasetApi(project_id)
         self._execution_api = execution_api.ExecutionsApi(project_id)
         self._log = logging.getLogger(__name__)
 
-    def download_logs(self, execution):
+    def download_logs(self, execution, path=None):
         """Download execution logs to current directory
+        :param path: path to download the logs
```
SirOibaf, line 34: Keep the documentation sorted please, `path` is the last parameter.

python/hopsworks/execution.py
```diff
@@ -178,7 +178,7 @@ def success(self):
                 return True
         return None
 
-    def download_logs(self):
+    def download_logs(self, path=None):
```
SirOibaf, line 181: This is the user facing method, here the `path` parameter is not documented.

SirOibaf (changes requested):
Can you also add a workflow test for this please?

SirOibaf:
@gibchikafa - Did you see the test failing because of the circular dependency? Can you please fix it as well?

SirOibaf (approved):

Issue #184, opened 2023-11-09
Hopsworks prints API Key when configured incorrectly


MKLepium:
If the API Key is incorrectly printed as a secret in (for example huggingface.co) and contains a "[API-Key]\n", newline at the end you the resulting stacktrace will contain the plain API Key. The http 

![image](https://github.com/logicalclocks/hopsworks-api/assets/52952859/3b050345-682d-4090-8549-e717e21c3d2f)

I have redacted my API key, but this error is only there when the API Key contains a \n at the end. If I remove the \n from the end of my API key, the error disappears.

MKLepium:
One way to test for this case would be:


    def test_login_newline_in_api_key(self):
        try:
            imaginaryApiKey = "ImaginaryApiKey\n"
            project = hopsworks.login(api_key_value=imaginaryApiKey)
        except Exception as e:
            self.assertNotIn(imaginaryApiKey.strip(), str(e))


If run with:

`python -m unittest tests.hopsworks.test_login.TestLogin.test_login_newline_in_api_key`

Currently the output would be:

```
Connected. Call `.close()` to terminate connection gracefully.
InsecureRequestWarning: Unverified HTTPS request is being made to host 'c.app.hopsworks.ai'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
F
======================================================================
FAIL: test_login_newline_in_api_key (tests.hopsworks.test_login.TestLogin)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mk/hopsworks-api/python/tests/hopsworks/test_login.py", line 236, in test_login_newline_in_api_key
    project = hopsworks.login(api_key_value=imaginaryApiKey)
ValueError: Invalid header value b'ApiKey ImaginaryApiKey\n'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/mk/hopsworks-api/python/tests/hopsworks/test_login.py", line 239, in test_login_newline_in_api_key
    self.assertNotIn(imaginaryApiKey.strip(), str(e))
AssertionError: 'ImaginaryApiKey' unexpectedly found in "Invalid header value b'ApiKey ImaginaryApiKey\\n'"

----------------------------------------------------------------------
Ran 1 test in 0.240s

FAILED (failures=1)
```

PR #185, opened 2023-11-21
added testcase for newline in api key

MKLepium:
Addition to what is written in [#184](https://github.com/logicalclocks/hopsworks-api/issues/184)

The current implementation of the authorization header results in an edge case where, when the resulting header is deemed invalid from the underlying HTTP library (urllib3), a value error is thrown, which is simply printed. This includes the content of the authorization header (the API key).

By simply stripping any leading or trailing white spaces (which include newlines), this should get rid of the error.

I can also not think of any edge cases where stripping the API key might lead to problems, since only whitespace leading or trailing the actual API key will be removed.

One could make the case that the strip is unnecessary for the bearer token, but my argument for keeping it is that it

1. can not hurt to have it
2. the time complexity of stripping the string once can be disregarded.

Last point:
Why did I choose strip instead of strip? If we believe that a user, while pasting the API key, might make the mistake of entering a newline at the end of the API key, he might also paste a newline at the beginning.

### DIFF
--- a/python/hopsworks/client/auth.py
+++ b/python/hopsworks/client/auth.py
@@ -24,7 +24,7 @@ class BearerAuth(requests.auth.AuthBase):
         self._token = token
 
     def __call__(self, r):
-        r.headers["Authorization"] = "Bearer " + self._token
+        r.headers["Authorization"] = "Bearer " + self._token.strip()
         return r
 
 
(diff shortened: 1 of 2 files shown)


SirOibaf (approved):
Thanks

PR #191, opened 2024-02-05
[HWORKS-985] Handle transient errors in Execution.download_logs() with retry

robzor92:


python/hopsworks/engine/execution_engine.py
```diff
@@ -52,21 +52,51 @@ def download_logs(self, execution, path=None):
...
+
         out_path = None
         if execution.stdout_path is not None and self._dataset_api.exists(
             execution.stdout_path
         ):
-            out_path = self._dataset_api.download(
-                execution.stdout_path, download_log_dir
-            )
-
+            while retries < max_num_retries:
```
ErmiasG, line 62: Can we make this a method like downloadLog(path)

ErmiasG (approved):

ErmiasG (approved):

PR #192, opened 2024-03-21
[FSTORE-1320] Fix hopsworks.login second attempt not taking parameters into account

vatj:
Cleanup client irrespective of whether connection was set properly


vatj:
Both adding client.stop in logout and in connect method of connection are sufficient to fix the issue. It does feel like both make sense to have though

robzor92 (approved):

PR #202, opened 2024-05-07
[FSTORE-1389] get_secrets_api and create_project should be accessible as hopsworks module functions

robzor92:


vatj (approved):
Looks good. Maybe you can add a show_project_list to hopsworks module? In the spirit of not having to go check in the UI

PR #213, opened 2024-07-02
[HWORKS-1048] Support multiple modularized project environments

robzor92:


python/hopsworks/core/environment_api.py
```diff
@@ -29,7 +32,7 @@ def __init__(
 
         self._environment_engine = environment_engine.EnvironmentEngine(project_id)
 
-    def create_environment(self, await_creation=True):
+    def create_environment(self, name: str, description: Optional[str] = None, base_environment_name: Optional[str] = "python-feature-pipeline", await_creation: Optional[bool] = True) -> environment.Environment:
```
javierdlrm, line 35: If the `base_environment_name` is the name of the image used as base for this environment, we can provide an Enum class to help users pick the correct names for the base images.

python/hopsworks/core/environment_api.py
```diff
@@ -99,30 +110,37 @@ def get_environment(self):
 
         env_api = project.get_environment_api()
 
-        env = env_api.get_environment()
+        env = env_api.get_environment("my_custom_environment")
 
         ```
         # Returns
```
javierdlrm, line 116: `# Arguments` is missing with the description of the new paramenter.

python/hopsworks/core/environment_api.py
```diff
@@ -99,30 +110,37 @@ def get_environment(self):
...
+        headers = {"content-type": "application/json"}
+        return environment.Environment.from_response_json(
+            _client._send_request(
+                "GET", path_params, query_params=query_params, headers=headers
+            ),
+            self._project_id,
+            self._project_name,
+        )
+
+    def _delete(self, name):
```
javierdlrm, line 134: Is this method "not exposed" on purpose? or should it be `def delete(self, name)` without underscore?
Can you add `#Arguments` with the description of the name parameter?

python/hopsworks/core/library_api.py
```diff
@@ -28,7 +28,7 @@ def __init__(
         self._project_id = project_id
         self._project_name = project_name
 
-    def install(self, library_name: str, python_version: str, library_spec: dict):
+    def install(self, library_name: str, name: str, library_spec: dict):
         """Create Python environment for the project"""
```
javierdlrm, line 32: Same about `#Arguments`. It guess it can raise `RestAPIError` as well.

python/hopsworks/core/environment_api.py
```diff
@@ -57,21 +63,26 @@ def create_environment(self, await_creation=True):
             self._project_id,
             "python",
             "environments",
-            client.get_python_version(),
+            name,
         ]
         headers = {"content-type": "application/json"}
+        data = {"name": name,
+                "baseImage": {
+                    "name": base_environment_name,
```
javierdlrm, line 71: If you add an Enum for the available base environment names, we can validate the name before sending the request to the backend.

python/hopsworks/environment.py
```diff
@@ -77,7 +82,17 @@ def python_version(self):
...
+    def name(self):
+        """Name of the environment"""
+        return self._name
+
+    @property
+    def description(self):
+        """Description of the environment"""
+        return self._description
+
+    def install_wheel(self, path, await_installation: Optional[bool] = True):
```
javierdlrm, line 95:
```suggestion
    def install_wheel(self, path: str, await_installation: Optional[bool] = True):
```

python/hopsworks/environment.py
```diff
@@ -116,15 +132,15 @@ def install_wheel(self, path, await_installation=True):
...
         )
 
         if await_installation:
-            return self._environment_engine.await_library_command(library_name)
+            return self._environment_engine.await_library_command(self.name, library_name)
 
         return library_rest
 
-    def install_requirements(self, path, await_installation=True):
+    def install_requirements(self, path, await_installation: Optional[bool] = True):
```
javierdlrm, line 143:
```suggestion
    def install_requirements(self, path: str, await_installation: Optional[bool] = True):
```

python/pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
 
 dependencies = [
     "hsfs[python] @ git+https://git@github.com/logicalclocks/feature-store-api@master#subdirectory=python",
-    "hsml @ git+https://git@github.com/logicalclocks/machine-learning-api@main#subdirectory=python",
+    "hsml @ git+https://git@github.com/robzor92/machine-learning-api@HWORKS-1048#subdirectory=python",
```
javierdlrm, line 26: Just a reminder, I know you are aware of this.

javierdlrm (changes requested):

python/hopsworks/core/environment_api.py
robzor92: Yes the _delete is hidden by design. The delete operation is done on the environment object itself.

python/hopsworks/core/environment_api.py
robzor92: resolved

python/hopsworks/core/library_api.py
robzor92: This function should actually be hidden, it's only used internally by Environment.install_wheel() etc... But I added the docs string.

python/hopsworks/core/environment_api.py
robzor92: The backend validates the name and sends a list of valid environment names back. I think adding the enum will come at a later stage maybe when I'm back.

python/hopsworks/core/environment_api.py
robzor92: The backend validates the name and sends a list of valid environment names back. I think adding the enum will come at a later stage maybe when I'm back.

python/pyproject.toml
robzor92: 👍

python/hopsworks/environment.py
robzor92: fixed

python/hopsworks/environment.py
robzor92: fixed

javierdlrm (changes requested):
Looks good to me. I'll approve once the github repo is changed back to logicalclocks.

javierdlrm (approved):

PR #219, opened 2024-07-03
[FSTORE-1454] Unify variable_api and project_api

aversey:


python/hopsworks/core/variable_api.py
```diff
@@ -43,3 +66,75 @@ def get_version(self, software: str):
...
+        """
+
+        _client = client.get_instance()
+        path_params = [
+            "variables",
+            "enable_flyingduck",
+        ]
+
+        resp = _client._send_request("GET", path_params)
+        return resp["successMessage"] == "true"
```
bubriks, line 102: just as i minor improvement i think we could do something like this (also applies for other methods in this class):
```suggestion
    def get_flyingduck_enabled(self) -> bool:
        """Check if Flying Duck is enabled on the backend.

        # Returns
            `True`: If flying duck is availalbe, `False` otherwise.
        # Raises
            `RestAPIError`: If unable to obtain the flag's value.
        """

        return self.get_variable("enable_flyingduck") == "true"
```

bubriks (approved):

python/hopsworks/core/variable_api.py
bubriks: but its ok if we fix it later as well, since it isn't crucial right now.

aversey: Sounds nice, thanks!

bubriks (approved):

kennethmhc (approved):

PR #228, opened 2024-07-15
 [FSTORE-1446] Merge GitHub Actions

aversey:


aversey:
For some reason, hsml.python is available only as python.hsml.python during docgen and in manual tests. [FSTORE-1473]

aversey:
The problem with hsml.python import is only present if python is started from the repo root directory.

kennethmhc (approved):

PR #229, opened 2024-07-16
[FSTORE-1453] Move client, decorators, variable_api and constants to hopsworks_common

aversey:


python/hopsworks/connection.py
```diff
@@ -271,7 +265,6 @@ def connect(self):
         )
 
         self._check_compatibility()
-        self._set_client_variables()
```
kennethmhc, line 274: This is missing after refactoring

kennethmhc (changes requested):

python/hopsworks/connection.py
aversey: I think it should be removed. It was setting only setting python version, but this is a legacy variable which is not used anymore (it was introduced [here](https://github.com/aversey/hopsworks-api/commit/96c6aab9484419c2a94475168156f873ddc20ab6), and recently remove [here](https://github.com/aversey/hopsworks-api/commit/f21ac7ed316c5c3fbcd76761fc06ab987eae0eb2))

kennethmhc (approved):

PR #239, opened 2024-07-23
Deduplicate constants

aversey:
JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hopsworks_common/constants.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+class Default:
+    """An object of this class is used in place of optional arguments in
+    cases when passing None should result in an exception or other behaviour.
+    Overwritting None with a default value is undesirable in some cases,
+    and this class solves the problem."""
+
+    def __repr__(self):
+        return "DEFAULT"
```
javierdlrm, line 25: Do we need to overwrite `__eq__` to check only for the instance type? Something like:
```
def __eq__(self, other):
    return isinstance(other, Default)
```
to use it like:
```
def method(self, optional=DEFAULT):
    if optional == DEFAULT:
        # use default value
```
or will DEFAULT return always the same object, and therefore match the same object ID?

aversey: DEFAULT returns the same object, but I liked your proposal, so it was implemented.

javierdlrm (approved):

PR #242, opened 2024-07-25
[FSTORE-1424] Feature logging for spark

kennethmhc:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/feature_view_engine.py
+++ b/python/hsfs/core/feature_view_engine.py
@@ -1110,7 +1110,8 @@ class FeatureViewEngine:
 
     def materialize_feature_logs(self, fv, wait, transform):
         if transform is None:
-            jobs = [self._get_logging_fg(fv, True).materialization_job, self._get_logging_fg(fv, False).materialization_job]
+            jobs = [self._get_logging_fg(fv, True).materialization_job,
+                    self._get_logging_fg(fv, False).materialization_job]
         else:
             jobs = [self._get_logging_fg(fv, transform).materialization_job]
         for job in jobs:
--- a/python/hsfs/core/feature_view_engine.py
+++ b/python/hsfs/core/feature_view_engine.py
@@ -17,7 +17,10 @@ from __future__ import annotations
 
 import datetime
 import warnings
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Union, TypeVar, Any
+
+import numpy as np
+import pandas as pd
 
 from hsfs import (
     client,
@@ -40,6 +43,7 @@ from hsfs.core import (
     training_dataset_engine,
 )
 from hsfs.core.feature_logging import FeatureLogging
+from hsfs.feature_view import FeatureView
 from hsfs.training_dataset_split import TrainingDatasetSplit
 
 
@@ -967,20 +971,19 @@ class FeatureViewEngine:
         )
 
     def _get_logging_fg(self, fv, transformed):
-        feature_logging = self.get_feature_logging(fv)
-        if transformed:
-            return feature_logging.transformed_features
-        else:
-            return feature_logging.untransformed_features
+        return self.get_feature_logging(fv).get_feature_group(transformed)
 
     def log_features(
         self,
-        fv,
-        features,
-        prediction=None,
-        transformed=False,
-        write_options=None,
-        training_dataset_version=None,
+        fv: FeatureView,
+        feature_logging: FeatureLogging,
+        features_rows: Union[
+            pd.DataFrame, list[list], np.ndarray, TypeVar("pyspark.sql.DataFrame")
+        ],
+        predictions: Optional[Union[pd.DataFrame, list[list], np.ndarray]] = None,
+        transformed: Optional[bool] = False,
+        write_options: Optional[Dict[str, Any]] = None,
+        training_dataset_version: Optional[int] = None,
         hsml_model=None,
     ):
         default_write_options = {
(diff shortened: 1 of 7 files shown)


python/hsfs/engine/python.py
```diff
@@ -734,20 +734,28 @@ def parse_schema_feature_group(
...
-        for feat_name in arrow_schema.names:
+        for i in range(len(arrow_schema.names)):
+            feat_name = arrow_schema.names[i]
             name = util.autofix_feature_name(feat_name)
             try:
-                converted_type = convert_pandas_dtype_to_offline_type(
-                    arrow_schema.field(feat_name).type
-                )
+                pd_type = arrow_schema.field(feat_name).type
+                if pd_type == "null" and feature_type_map.get(name):
```
javierdlrm, line 755: @manu-sj was using `pa.types.is_null` to check for null types. I think that's a safer option. ([link](https://github.com/logicalclocks/hopsworks-api/blob/b510e308170c6282de85dbb8124d98e4f499549c/python/hsfs/engine/python.py#L541))

python/hsfs/engine/python.py
```diff
@@ -1440,40 +1448,40 @@ def _convert_feature_log_to_df(feature_log, cols):
...
-        import uuid
-
         features = Engine._convert_feature_log_to_df(
-            features, [f.name for f in fg_features]
+            features, td_features
         )
         if td_predictions:
-            prediction = Engine._convert_feature_log_to_df(
-                prediction, [f.name for f in td_predictions]
+            predictions = Engine._convert_feature_log_to_df(
```
javierdlrm, line 1473: Do we need `Engine.` here? I see it removed in the next line for `cast_column_to_offline_type`

python/hsfs/feature_view.py
```diff
@@ -3756,12 +3782,16 @@ def materialize_log(self, wait: Optional[bool] = False):
...
+
+
         # Raises
             `hsfs.client.exceptions.RestAPIError` in case the backend fails to materialize the log.
         """
-        return self._feature_view_engine.materialize_feature_logs(self, wait)
+        return self._feature_view_engine.materialize_feature_logs(self, wait, transformed)
 
-    def delete_log(self, transformed: Optional[bool] = None):
+    def delete_log(self, transformed: Optional[bool] = None) -> None:
```
javierdlrm, line 3794: I don't think  `-> None` is needed, is it?

python/hsfs/feature_view.py
```diff
@@ -3773,10 +3803,13 @@ def delete_log(self, transformed: Optional[bool] = None):
             feature_view.delete_log()
             ```
 
-        # Raises
-            `hsfs.client.exceptions.RestAPIError` in case the backend fails to delete the log.
+         # Raises
+             `hsfs.client.exceptions.RestAPIError` in case the backend fails to delete the log.
```
javierdlrm, line 3807: Is this the correct indentation?

javierdlrm (changes requested):

python/hsfs/feature_view.py
```diff
@@ -3590,28 +3593,36 @@ def enable_logging(self) -> None:
...
+            pd.DataFrame, list[list], np.ndarray, TypeVar("pyspark.sql.DataFrame")
+        ],
         predictions: Optional[Union[pd.DataFrame, list[list], np.ndarray]] = None,
         transformed: Optional[bool] = False,
         write_options: Optional[Dict[str, Any]] = None,
         training_dataset_version: Optional[int] = None,
-        hsml_model=None,
-    ):
-        """Log features and optionally predictions for the current feature view.
+        hopsworks_model=None,
```
javierdlrm, line 3609: Do we need to specify `hopsworks_`? We could use just `model`, and `if not isinstance(model, hopsworks.model.Model)` we raise a ValueError("Model must be a hopsworks model").
Then the API looks cleaner

python/hsfs/feature_view.py
```diff
@@ -3632,22 +3643,23 @@ def log(
...
             self,
+            self.feature_logging,
             features,
             predictions,
             transformed,
             write_options,
             training_dataset_version=(
                 training_dataset_version or self.get_last_accessed_training_dataset()
             ),
-            hsml_model=hsml_model,
```
javierdlrm, line 3642: Another improvement can be to check if `training_dataset_version or self.get_last_accessed_training_dataset()` is still None, and if a model is provided, we get the training dataset version using provenance.
Then the training dataset version is optional if the model is passed, so simpler and less arguments.

python/hsfs/core/feature_view_engine.py
```diff
@@ -967,39 +970,46 @@ def get_feature_logging(self, fv):
...
+        fv: FeatureView,
+        feature_logging: FeatureLogging,
+        features_rows: Union[
+            pd.DataFrame, list[list], np.ndarray, TypeVar("pyspark.sql.DataFrame")
+        ],
+        predictions: Optional[Union[pd.DataFrame, list[list], np.ndarray]] = None,
+        transformed: Optional[bool] = False,
+        write_options: Optional[Dict[str, Any]] = None,
+        training_dataset_version: Optional[int] = None,
         hsml_model=None,
```
javierdlrm, line 986: Can you please change the name of `hsml_model` here as well? so we keep the same name across methods.

javierdlrm (changes requested):

python/hsfs/engine/python.py
kennethmhc: yes because `get_feature_logging_df` is static method

python/hsfs/feature_view.py
kennethmhc: why not?

python/hsfs/core/feature_view_engine.py
kennethmhc: I think we can keep the internal method as `hsml_model` so to make the variable more understandable

python/hsfs/feature_view.py
javierdlrm: Isn't it the default?

javierdlrm (approved):

PR #266, opened 2024-08-08
[FSTORE-1460] loadbalancer per service - Big breaking change

vatj:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


java/hsfs/src/main/java/com/logicalclocks/hsfs/engine/VectorServer.java
```diff
@@ -383,7 +384,7 @@ private void setupJdbcConnection(FeatureStoreBase featureStoreBase, Boolean exte
       // if external is true, replace the IP coming from the storage connector with the host
       // used during the connection setup
       String host;
-      Optional<Variable> loadbalancerVariable = variablesApi.get(VariablesApi.LOADBALANCER_EXTERNAL_DOMAIN);
+      Optional<Variable> loadbalancerVariable = variablesApi.get(VariablesApi.LOADBALANCER_EXTERNAL_DOMAIN_MYSQL);
```
robzor92, line 387: I guess the fallback here that assumes mysql is running on the head node is not correct anymore in the kubernetes world since we don't have a head node abstraction and services are in pods that may or may not run on a given host. Not sure what the correct fallback is here.

python/hopsworks_common/core/variable_api.py
```diff
@@ -94,16 +100,23 @@ def get_flyingduck_enabled(self) -> bool:
         """
         return self.get_variable("enable_flyingduck") == "true"
 
-    def get_loadbalancer_external_domain(self) -> str:
-        """Get domain of external loadbalancer.
+    def get_loadbalancer_external_domain(self, service: str) -> str:
+        """Get domain of external loadbalancer for MySQLd.
```
robzor92, line 104:
```suggestion
        """Get domain of external loadbalancer for a service.
```

python/hopsworks_common/core/variable_api.py
```diff
@@ -94,16 +100,23 @@ def get_flyingduck_enabled(self) -> bool:
...
 
         # Returns
-            `str`: The domain of external loadbalancer, if it is set up, otherwise empty string `""`.
+            `str`: The domain of external loadbalancer for MySQLd, if it is set up, otherwise empty string `""`.
```
robzor92, line 107:
```suggestion
            `str`: The domain of external loadbalancer for a service, if it is set up, otherwise empty string `""`.
```

python/hopsworks_common/core/variable_api.py
robzor92: Since my understanding is this function will be used for all services with loadbalancers and not just MySQL

robzor92 (changes requested):

java/hsfs/src/main/java/com/logicalclocks/hsfs/engine/VectorServer.java
vatj: There is no fallback. I will raise an exception for it.

PR #267, opened 2024-08-09
[FSTORE-1497] Draft schedule materialisation job on first insertion

vatj:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


SirOibaf:
Just 2 comments here:
- The scheduler waits for any ongoing executions before submitting a new execution. This to avoid running parallel inserts in the offline feature store, which is not something we support yet. However the Python code right now does not check if there is an ongoing execution for the same materialization job before triggering the execution. Having the materialization job scheduled in the background and doing insertion at the same time will cause issues if they end up being executed concurrently.

- Is there a specific reason we need this to be enabled by default? I would say that having an easier way to schedule the materialization job is a good idea if you are writing in streaming, but if you are writing with Python, the materialization job is kicked off already after the data has been sent to Kafka.

manu-sj (approved):

PR #286, opened 2024-08-19
[FSTORE-1476] Deduplicate connection

aversey:
JIRA Issue: FSTORE-1476

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [x] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hopsworks_common/connection.py
```diff
@@ -359,8 +432,13 @@ def close(self) -> None:
             conn.close()
             ```
         """
+        from hsfs import engine
+
         OpenSearchClientSingleton().close()
         client.stop()
+        engine.stop()
+        self._feature_store_api = None
+        self._model_api = None
```
vatj, line 441: model_serving_api, model_registry_api?

python/hopsworks_common/connection.py
```diff
@@ -319,6 +377,21 @@ def connect(self) -> None:
                     hostname_verification=self._hostname_verification,
                 )
 
+            from hsfs import engine
+            from hsfs.core import feature_store_api
+
+            # init engine
```
vatj, line 383: Why is this here and not when getting the feature store?

vatj (changes requested):
Move the engine init to the get_feature_store. We don't need it until then.

Check the model serving stuff

python/hopsworks_common/connection.py
aversey: It was this way in hsml before the merge: https://github.com/logicalclocks/machine-learning-api/blob/main/python/hsml/connection.py#L200

Also, I would prefer to remove `_model_api = None` instead of adding them.

vatj (approved):

PR #296, opened 2024-08-26
[FSTORE-471] Allow transformation functions for label features

manu-sj:
This PR adds support for adding transformation function to labels in the feature view.
 
**Changes done**
- Added property `transformation_labels` which provide names of labels after transformation functions have been applied to it.
- Added function `get_training_dataset_schema` which provide the schema of the features for a specific training dataset version.  Training dataset schema can only be generated after creation of a training dataset because the schema can vary based on training dataset statistics if the transformation `one_hot_encoder` is applied.

JIRA Issue: https://hopsworks.atlassian.net/browse/FSTORE-471

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [x] Unit Tests
- [x] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/feature_view_engine.py
+++ b/python/hsfs/core/feature_view_engine.py
@@ -189,7 +189,9 @@ class FeatureViewEngine:
         else:
             return self._feature_view_api.delete_by_name(name)
 
-    def get_schema(self, feature_view: FeatureView, training_dataset_version: int):
+    def get_training_dataset_schema(
+        self, feature_view: FeatureView, training_dataset_version: int
+    ):
         """
         Function that returns the schema of the training dataset generated using the feature view.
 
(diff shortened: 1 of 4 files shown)


o-alex (approved):

python/hsfs/feature_view.py
```diff
@@ -3844,6 +3844,29 @@ def to_dict(self) -> Dict[str, Any]:
...
+        """
+        Function that returns the schema of the training dataset that is generated from a feature view.
+        It provides the schema of the features after all transformation functions have been applied.
+
+        # Arguments
+            training_dataset_version: Version of the training dataset for which the schema is to be generated.
+
+        # Example
+            ```python
+            schema = feature_view.get_schema(training_dataset_version=1)
```
bubriks, line 3859:
```suggestion
            schema = feature_view.get_training_dataset_schema(training_dataset_version=1)
```

python/hsfs/core/feature_view_engine.py
```diff
@@ -183,6 +189,93 @@ def delete(self, name, version=None):
...
+        if not feature_view.transformation_functions:
+            # Return the features in the feature view if the no transformation functions in the feature view.
+            return feature_view.features
+        else:
+            # Create schema based on transformation function.
+            (
+                transformed_feature_names,
+                output_types_features,
+                transformed_label_names,
+                output_types_labels,
```
bubriks, line 216: no sure i got why do it like this?
wouldn't it be the same using something like this:
```
transformed_names = []
output_types = []
```
Aka. no separating features and lables, since they are joined either way?
maybe even just having list of `training_dataset_feature.TrainingDatasetFeature` and adding to it directly?

bubriks (changes requested):

python/hsfs/feature_view.py
manu-sj: Fixed.

python/hsfs/core/feature_view_engine.py
manu-sj: Yes you I right I have updated the code based on the suggestion.

bubriks: why not just work with a list of `training_dataset_feature.TrainingDatasetFeature`? i think it would be cleaner, right now you have `transformed_features` and `transformed_labels`, this make you have multiple loops to construct list of `training_dataset_feature.TrainingDatasetFeature`.

manu-sj: The schema of the transformed feature follows an order. Any training dataset or data read from the feature view will contains all the untransformed_feature followed by the transformed_features. 
So the `training_dataset_schema` must return a list that has features in the order  untransformed_featrures, transformed_features, labels and transformed_labels. 

To preserve the order, I think if we use single list we would need two for loops one for labels and the other for normal features. That was why I did the changes using two lists instead of two for loops, but do you think its better and cleaner to use two for loops instead of two lists.

bubriks (approved):

PR #299, opened 2024-08-26
Add telemetry to hsml and hopsworks, move usage to hopsworks_common

aversey:
JIRA Issue: FSTORE-1455

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


README.md
```diff
@@ -164,6 +164,23 @@ data = { "instances": [ model.input_example ] }
 predictions = deployment.predict(data)
 ```
 
+## Usage
+
+Usage data is collected for improving quality of the library.
+It is turned on by default if the backend is "c.app.hopsworks.ai".
```
vatj, line 170: Call it by its name and hide the host name as a link to app.

python/hopsworks/usage.py
```diff
@@ -0,0 +1,38 @@
+#
+#   Copyright 2024 Hopsworks AB
```
vatj, line 2: Not needed there was no usage in hopsworks before

python/hsml/deployment.py
```diff
@@ -32,6 +33,7 @@
 class Deployment:
     """Metadata object representing a deployment in Model Serving."""
 
+    @usage.method_logger
```
vatj, line 36: Don't need to be on the init I would say

python/hsml/deployment.py
```diff
@@ -165,6 +171,7 @@ def is_stopped(self, or_created=True) -> bool:
             )
         )
 
+    @usage.method_logger
```
vatj, line 174: Can this add a perf cost?

python/hsml/model_schema.py
```diff
@@ -31,6 +32,7 @@ class ModelSchema:
         `ModelSchema`. The model schema object.
     """
 
+    @usage.method_logger
```
vatj, line 35: Don't think it is needed here either

python/hsml/model.py
```diff
@@ -39,6 +40,7 @@
 class Model:
     """Metadata object representing a model in the Model Registry."""
 
+    @usage.method_logger
```
vatj, line 43: Don't think we need to log this one either

python/hsml/schema.py
```diff
@@ -33,6 +34,7 @@ class Schema:
         `Schema`. The schema object.
     """
 
+    @usage.method_logger
```
vatj, line 37: Unnecessary

vatj (changes requested):
There is a confusion I think on what the logger should do. We are trying to understand how people use the library not what values are used to create a model.

Additionally perf critical method such as predict should not be wrapped by the decorator I would say

README.md
aversey: I am not sure if I understand your comment. Do you want me to replace `"c.app.hopsworks.ai"` with `hopsworks.constants.HOSTS.APP_HOST`?

python/hopsworks_common/core/project_api.py
```diff
@@ -113,6 +113,7 @@ def _create_project(
         print("Project created successfully, explore it at " + project.get_url())
         return project
 
+    @usage.method_logger
     def get_client(self):
```
javierdlrm, line 117: There is no `get_project_api()` method available, so I don't think users can call this method directly. We can skip the usage logger here.

Moreover, I can't find where this method is used. Can you double-check?

python/hopsworks_common/core/services_api.py
```diff
@@ -18,10 +18,11 @@
 
 from typing import Any, Dict
 
-from hopsworks_common import client
+from hopsworks_common import client, usage
 
 
 class ServicesApi:
+    @usage.method_logger
     def get_service(self, service: str) -> Dict[str, Any]:
```
javierdlrm, line 26: Same here. I think we can remove the usage logger, and also I can't find where this method is used or how users can access it without instantiating the ServiceApi class by themselves.

javierdlrm (changes requested):

javierdlrm (approved):

vatj (approved):

PR #315, opened 2024-09-04
[FSTORE-1530] Training Data Materialization using Arrowflight fails with PermissionError when a workflow test run as a Kubernetes Job

manu-sj:
This PR fixes a issue in which training dataset materialization using Arrowflight was throwing a Permission Error

**Root Cause**
When trying to creating training dataset using Arrowflight a request is send from the python client to arrowflight that contains the feature store name along with other parameters. The feature store name is by default cast to into lower case letters before sending. 
Arrowflight then uses this feature store name to construct the path required to access the training dataset directory but the constructed path is incorrect because the feature store name has been casted to lower cases. 
So Arrowflight requests lock access for the parent directories (/Projectto create the provided path which gets rejected because it should not be able to access the directory.

**Fix Done**
Send the unformatted project name to the Arrowflight client during project creation.

JIRA Issue: https://hopsworks.atlassian.net/browse/FSTORE-1530

Priority for Review: -

Related PRs: https://github.com/logicalclocks/flyingduck/pull/83

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/arrow_flight_client.py
+++ b/python/hsfs/core/arrow_flight_client.py
@@ -29,7 +29,7 @@ import pyarrow._flight
 import pyarrow.flight
 from hopsworks_common import client
 from hopsworks_common.client.exceptions import FeatureStoreException
-from hsfs import feature_group, util
+from hsfs import feature_group
 from hsfs.constructor import query
 from hsfs.core.variable_api import VariableApi
 from hsfs.storage_connector import StorageConnector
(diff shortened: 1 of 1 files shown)


python/hsfs/core/arrow_flight_client.py
```diff
@@ -444,9 +444,7 @@ def create_training_dataset(
         self, feature_view_obj, training_dataset_obj, query_obj, arrow_flight_config
     ):
         training_dataset = {}
-        training_dataset["fs_name"] = util.strip_feature_store_suffix(
-            training_dataset_obj.feature_store_name
-        )
+        training_dataset["project_name"] = self._client._project_name
```
SirOibaf, line 447: I think it's fine here, because I don't think it's possible to create a training dataset in a shared feature store. 
But as a general note, be careful when you do this type of things `self._client._project_name` refers to the project the user is connected to, not the feature store they are working in.

So for example, if you do:
```
project = hopsworks.login(project="source_project")
fs = project.get_feature_store(name="shared_feature_store")
fv = fs.get_feature_view(...)
fv.create_training_dataset()
```

then your code above will be using `source_project` and not `shared_feature_store` project.

SirOibaf (approved):

PR #323, opened 2024-09-10
[FSTORE-1533] Importing hopsworks fails due to core dump while importing polars

manu-sj:
This PR fixes an issue in which hopworks is not importable because Polars throws a "" due to AVX instructions being unavailable.

https://github.com/pola-rs/polars?tab=readme-ov-file#legacy 

Fix Done: 
Make polars a optional dependency. The changes being done resembles the changes done for 3.8 (https://github.com/logicalclocks/feature-store-api/pull/1379)

JIRA Issue: https://hopsworks.atlassian.net/browse/FSTORE-1533

Related PRs: 
https://github.com/logicalclocks/feature-store-api/pull/1379
https://github.com/logicalclocks/loadtest/pull/459
https://github.com/logicalclocks/docker-images/pull/437

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/python/hopsworks_common/core/constants.py
+++ b/python/hopsworks_common/core/constants.py
@@ -49,9 +49,9 @@ HAS_NUMPY: bool = importlib.util.find_spec("numpy") is not None
 HAS_POLARS: bool = importlib.util.find_spec("polars") is not None
 polars_not_installed_message = (
     "Polars package not found. "
-    "If you want to use Polars with Hopsworks you can install the corresponding extras "
-    """`pip install hopsworks[polars]` or `pip install "hopsworks[polars]"` if using zsh. """
-    "You can also install polars directly in your environment e.g `pip install polars`. "
+    "If you want to use Polars with Hopsworks you can install the corresponding extra via "
+    """'`pip install "hopsworks[polars]"`. '"""
+    "You can also install polars directly in your environment with `pip install polars`. "
     "You will need to restart your kernel if applicable."
 )
 
--- a/python/hopsworks_common/core/constants.py
+++ b/python/hopsworks_common/core/constants.py
@@ -47,6 +47,13 @@ HAS_ARROW: bool = importlib.util.find_spec("pyarrow") is not None
 HAS_PANDAS: bool = importlib.util.find_spec("pandas") is not None
 HAS_NUMPY: bool = importlib.util.find_spec("numpy") is not None
 HAS_POLARS: bool = importlib.util.find_spec("polars") is not None
+polars_not_installed_message = (
+    "Polars package not found. "
+    "If you want to use Polars with Hopsworks you can install the corresponding extras "
+    """`pip install hopsworks[polars]` or `pip install "hopsworks[polars]"` if using zsh. """
+    "You can also install polars directly in your environment e.g `pip install polars`. "
+    "You will need to restart your kernel if applicable."
+)
 
 # SQL packages
 HAS_SQLALCHEMY: bool = importlib.util.find_spec("sqlalchemy") is not None
--- a/python/hopsworks_common/core/type_systems.py
+++ b/python/hopsworks_common/core/type_systems.py
@@ -26,8 +26,8 @@ from hopsworks_common.core.constants import (
     HAS_ARROW,
     HAS_PANDAS,
     HAS_POLARS,
-    polars_not_installed_message,
 )
+from hopsworks_common.decorators import uses_polars
 
 
 if TYPE_CHECKING:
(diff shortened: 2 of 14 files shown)


python/hopsworks_common/core/type_systems.py
```diff
@@ -233,6 +241,9 @@ def cast_column_to_offline_type(
     if isinstance(feature_column, pd.Series):
         return cast_pandas_column_to_offline_type(feature_column, offline_type.lower())
     elif isinstance(feature_column, pl.Series):
```
aversey, line 243: This line contains error: unknown pl.

python/hsfs/feature_view.py
```diff
@@ -76,16 +76,25 @@
...
+        pl.DataFrame,
+    ]
+else:
+    TrainingDatasetDataFrameTypes = Union[
+        pd.DataFrame,
+        TypeVar("pyspark.sql.DataFrame"),  # noqa: F821
+        TypeVar("pyspark.RDD"),  # noqa: F821
+        np.ndarray,
+        List[List[Any]],
+    ]
```
aversey, line 97: Change this to:
```
TrainingDatasetDataFrameTypes = Union[
    pd.DataFrame,
    TypeVar("pyspark.sql.DataFrame"),  # noqa: F821
    TypeVar("pyspark.RDD"),  # noqa: F821
    np.ndarray,
    List[List[Any]],
]
if HAS_POLARS:
    import polars as pl
    TrainingDatasetDataFrameTypes.append(pl.DataFrame)
```

aversey (changes requested):
Cool!

One general remark also: it is preferable for functions which require polars to raise the human-readable exception with polars_not_installed_message instead of raising "name not found error" (see https://github.com/logicalclocks/hopsworks-api/blob/e3786ede9e0fa947ef24f721a4de839f2542d4f1/python/hopsworks_common/decorators.py#L79 for a decorator which may be useful for this purpose). I will make a more thorough review later.

python/hsfs/feature_view.py
manu-sj: I dont think we can do an append on an Union type. That is why I had it as two separate declarations

image

python/hopsworks_common/core/type_systems.py
manu-sj: Fixed

manu-sj (commented):
Added `uses_polars` decorator for throwing an error if a function that only uses polars is called.  
The decorator is applied to the functions `cast_polars_column_to_offline_type` and `_read_polars` since they are only called with polars objects as input. 
In all other cases the same error `ModuleNotFoundError(polars_not_installed_message)` has been thrown

python/hsfs/feature_view.py
aversey: Oj, sorry. =) My bad. It should be:
```
TrainingDatasetDataFrameTypes = Union[TrainingDatasetDataFrameTypes, pl.DataFrame]
```

python/hopsworks_common/core/constants.py
```diff
@@ -47,6 +47,13 @@
 HAS_PANDAS: bool = importlib.util.find_spec("pandas") is not None
 HAS_NUMPY: bool = importlib.util.find_spec("numpy") is not None
 HAS_POLARS: bool = importlib.util.find_spec("polars") is not None
+polars_not_installed_message = (
+    "Polars package not found. "
+    "If you want to use Polars with Hopsworks you can install the corresponding extras "
+    """`pip install hopsworks[polars]` or `pip install "hopsworks[polars]"` if using zsh. """
```
aversey, line 53: I think it is better to simply say:
```
'`pip install "hopsworks[polars]"`. '
```
Since it works for all shells.

python/hopsworks_common/core/constants.py
```diff
@@ -47,6 +47,13 @@
 HAS_PANDAS: bool = importlib.util.find_spec("pandas") is not None
 HAS_NUMPY: bool = importlib.util.find_spec("numpy") is not None
 HAS_POLARS: bool = importlib.util.find_spec("polars") is not None
+polars_not_installed_message = (
+    "Polars package not found. "
+    "If you want to use Polars with Hopsworks you can install the corresponding extras "
```
aversey, line 52: Maybe it is better to say:
```
... install the corresponding extra via
```

python/hopsworks_common/core/constants.py
```diff
@@ -47,6 +47,13 @@
...
+    """`pip install hopsworks[polars]` or `pip install "hopsworks[polars]"` if using zsh. """
+    "You can also install polars directly in your environment e.g `pip install polars`. "
```
aversey, line 54: "e.g" kind of doesn't make sense for me here, maybe it is better to use "with".

aversey (dismissed):

vatj (approved):

PR #328, opened 2024-09-11
[FSTORE-1537] Managed feature group metadata changes

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


bubriks:
`hoodie.datasource.hive_sync.enable` is set to `True` meaning hive metastore entry will be created

python/hsfs/feature_group.py
```diff
@@ -2049,6 +2059,24 @@ def online_enabled(self) -> bool:
...
+
+    def get_uri(self) -> str:
+        """Location of data."""
+        if (self.storage_connector is None):
+            return self.location
+        else:
+            path = self.storage_connector._get_path(self.path)
+            if engine.get_type().startswith("spark"):
+                path = engine.get_instance().setup_storage_connector(self.storage_connector, path)
+            return path
```
kennethmhc, line 2078:
```suggestion
            return self.storage_connector._get_path(path)
```

python/hsfs/engine/spark.py
```diff
@@ -213,13 +220,16 @@ def register_hudi_temporary_table(
...
+
         hudi_engine_instance.register_temporary_table(
             hudi_fg_alias,
             read_options,
         )
-        hudi_engine_instance.reconcile_hudi_schema(
-            self.save_empty_dataframe, hudi_fg_alias, read_options
-        )
+
+        if (hudi_fg_alias._feature_group.storage_connector is None):
```
kennethmhc, line 229: I think `reconcile_hudi_schema` is needed even if there is storage_connector

python/hsfs/constructor/fs_query.py
```diff
@@ -99,16 +99,21 @@ def register_external(
...
+        if self._on_demand_fg_aliases is not None:
+            for external_fg_alias in self._on_demand_fg_aliases:
+                if type(external_fg_alias.on_demand_feature_group).__name__ == "SpineGroup":
+                    external_fg_alias.on_demand_feature_group.dataframe = spine
+                engine.get_instance().register_external_temporary_table(
+                    external_fg_alias.on_demand_feature_group,
+                    external_fg_alias.alias,
+                )
+
+        if self._hudi_cached_feature_groups is not None:
```
kennethmhc, line 111: why do you need to register hudi temp table here? is it not registered by `register_hudi_tables`?

kennethmhc (changes requested):

python/hsfs/feature_group.py
bubriks: we have to prepare spark for the storage connector.

python/hsfs/engine/spark.py
bubriks: we don't have the hive schema to compare against. (but i guess we could use FG features though)

python/hsfs/feature_group.py
kennethmhc: it should be done when you actually read data, right?

bubriks: yeah, but `prepare_spark` also updates the URI. It changes `s3://` to `s3a://` so i hoped to avoid doing additional checks everywhere, since it is needed not only for `read`, but also for operations `write`, `register_temporary_table`, `append_column`, and `delete_record`.

kennethmhc: ok, but then the method name can be more explicit. Also, we should return the path for python client as well.

bubriks: i updated name and now i use location received from the back-end. This means more testing will be needed and ill check if there are no issues to just return `s3a://` from hopsworks instead of `s3://` and then changing it in the api.

kennethmhc (approved):

PR #334, opened 2024-09-16
[HWORKS-1565][APPEND] Make sure the spark program exits after completion

manu-sj:
This PR fixes the issue in which exceptions are lost if a spark Job Fails.

Root Cause:
PR https://github.com/logicalclocks/hopsworks-api/pull/275 added a `system.exit(1)` to kill the Kubernetes pod in case of spark Job Failure. However this causes the exception logs to be lost since they are not re-raised after the finally block.

https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions

Fix Done: 
As a temporary fix the exceptions before calling System.exit(1). Jira ticket has been created for further investigation of the need for a System.exit

 
JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/utils/java/src/main/java/com/logicalclocks/utils/MainClass.java
+++ b/utils/java/src/main/java/com/logicalclocks/utils/MainClass.java
@@ -132,15 +132,6 @@ public class MainClass {
       }
       if (!success) {
         System.exit(1);
-        LOGGER.info("Closing spark session...");
-        try {
-          SparkEngine.getInstance().closeSparkSession();
-        } catch (Exception e) {
-          LOGGER.error("Error closing spark session", e);
-        }
-        if (!success) {
-          System.exit(1);
-        }
       }
     }
     System.exit(0);
--- a/utils/java/src/main/java/com/logicalclocks/utils/MainClass.java
+++ b/utils/java/src/main/java/com/logicalclocks/utils/MainClass.java
@@ -121,8 +121,7 @@ public class MainClass {
         SparkEngine.getInstance().streamToHudiTable(streamFeatureGroup, writeOptions);
       }
       success = true;
-    }
-    catch (Exception e) {
+    } catch (Exception e) {
       e.printStackTrace();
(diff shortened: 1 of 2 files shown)


utils/python/hsfs_utils.py
```diff
@@ -304,6 +305,9 @@ def parse_isoformat_date(da: str) -> datetime:
             run_feature_monitoring(job_conf)
 
         success = True
+    except Exception:
+        # Printing stack trace of exception so that logs are not lost.
+        print(traceback.format_exc())
```
robzor92, line 310: I have never used the traceback module. This was tested manually right?

utils/java/src/main/java/com/logicalclocks/utils/MainClass.java
```diff
@@ -130,6 +132,15 @@ public static void main(String[] args) throws Exception {
       }
       if (!success) {
         System.exit(1);
```
robzor92, line 134: Won't the System.exit(1); on line 134 prevent the added code below from running?

robzor92 (changes requested):

utils/java/src/main/java/com/logicalclocks/utils/MainClass.java
manu-sj: Sorry this was my mistake. There was a duplication of code I fixed it now.

utils/python/hsfs_utils.py
manu-sj: Yes I have tested it manually. I have attached the screenshot of the logs printed after the fix. Source :https://stackoverflow.com/questions/3702675/catch-and-print-full-python-exception-traceback-without-halting-exiting-the-prog
image

robzor92 (approved):

PR #347, opened 2024-10-01
[FSTORE-1536] Support delta table in python client

kennethmhc:
add DeltaCachedFeatureGroups

read delta s3 python

read write delta s3

add delta to dev

add delta

read delta on s3

This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/arrow_flight_client.py
+++ b/python/hsfs/core/arrow_flight_client.py
@@ -578,7 +578,7 @@ def _serialize_featuregroup_connector(fg, query, on_demand_fg_aliases):
         connector["type"] = fg.storage_connector.type
         connector["options"] = fg.storage_connector.connector_options()
         if fg.storage_connector.type == StorageConnector.S3:
-            connector["path"] = fg.location
+            connector["options"]["path"] = fg.location
         connector["query"] = ""
         if query._left_feature_group == fg:
             connector["filters"] = _serialize_filter_expression(
--- a/python/hsfs/core/arrow_flight_client.py
+++ b/python/hsfs/core/arrow_flight_client.py
@@ -124,7 +124,13 @@ def _is_query_supported_rec(query: query.Query):
         and query._left_feature_group.storage_connector.type
         in ArrowFlightClient.SUPPORTED_EXTERNAL_CONNECTORS
     )
-    supported = hudi_no_time_travel or supported_connector
+    delta_s3 = (
+        isinstance(query._left_feature_group, feature_group.FeatureGroup)
+        and query._left_feature_group.time_travel_format == "DELTA"
+        and query._left_feature_group.storage_connector
+        and query._left_feature_group.storage_connector.type == StorageConnector.S3
+    )
+    supported = hudi_no_time_travel or supported_connector or delta_s3
     for j in query._joins:
         supported &= _is_query_supported_rec(j._query)
     return supported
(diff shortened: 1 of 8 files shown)


python/hsfs/storage_connector.py
```diff
@@ -369,6 +369,16 @@ def prepare_spark(self, path: Optional[str] = None) -> Optional[str]:
...
         return engine.get_instance().setup_storage_connector(self, path)
 
+    def connector_options(self) -> Dict[str, Any]:
+        """Return options to be passed to an external S3 connector library"""
+        return {
+            "access_key": self.access_key,
+            "secret_key": self.secret_key,
+            "session_token": self.session_token,
+            "path": self.path,
+            "region": "us-east-2",
```
bubriks, line 379: region is user defined

python/hsfs/core/arrow_flight_client.py
```diff
@@ -566,8 +573,23 @@ def _serialize_featuregroup_connector(fg, query, on_demand_fg_aliases):
                     connector["filters"] = _serialize_filter_expression(
                         join_obj._query._filter, join_obj._query, True
                     )
+    elif fg.time_travel_format == "DELTA":
+        connector["time_travel_type"] = "delta"
+        connector["type"] = fg.storage_connector.type
+        connector["options"] = fg.storage_connector.connector_options()
```
bubriks, line 579: what if delta table is stored on hopsfs and not s3?

bubriks (changes requested):

python/hsfs/storage_connector.py
kennethmhc: fixed

python/hsfs/core/arrow_flight_client.py
kennethmhc: no, it only work for s3

python/hsfs/storage_connector.py
```diff
@@ -369,6 +369,16 @@ def prepare_spark(self, path: Optional[str] = None) -> Optional[str]:
         """
         return engine.get_instance().setup_storage_connector(self, path)
 
+    def connector_options(self) -> Dict[str, Any]:
+        """Return options to be passed to an external S3 connector library"""
+        return {
+            "access_key": self.access_key,
+            "secret_key": self.secret_key,
+            "session_token": self.session_token,
+            "path": self.path,
```
bubriks, line 378: what is this path? shouldn't it be received from feature group?

bubriks: for ex:
`fg.location` == 's3://ralfsbucket/fg_delta_1'

kennethmhc: right, it should be the location of the feature group

python/hsfs/core/arrow_flight_client.py
```diff
@@ -566,8 +573,25 @@ def _serialize_featuregroup_connector(fg, query, on_demand_fg_aliases):
...
+        if fg.storage_connector.type == StorageConnector.S3:
+            connector["path"] = fg.location
```
bubriks, line 581: shouldn't `path` be under `options`?

bubriks (changes requested):

python/hsfs/core/arrow_flight_client.py
kennethmhc: right!

bubriks: Actually, i think `path` should be on the same level as `query`. Since i believe it is either one or the other. Therefore, maybe adapting flying duck would be more consistent.

kennethmhc: It should be in connector level because you may join delta fg

bubriks (approved):

PR #405, opened 2024-11-19
[FSTORE-1606] Allow ´entries´ to be None while retrieving feature vectors from a feature view with only on-demand features

manu-sj:
This PR fixes an issue in which dummy data needs to be provided using `entires` to the function `get_feature_vector` while retrieving feature vectors from a feature view with only on-demand features. 

**Issue:**
User have to always provide `entires` in the function `get_feature_vector` when reading feature vectors from a feature view. 

In the use case in which a feature view contains only on-demand features,  all features are computed while retrieving feature vectors using provided request parameter. Hence `entries` is not a mandatory parameter.

**Fix Done:**
Made `entries` into an optional parameter in cases where the feature view contains only on-demand features.

JIRA Issue: https://hopsworks.atlassian.net/browse/FSTORE-1606

Priority for Review: -

Related PRs: https://github.com/logicalclocks/loadtest/pull/485

**How Has This Been Tested?**

- [ ] Unit Tests
- [x] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/vector_server.py
+++ b/python/hsfs/core/vector_server.py
@@ -149,6 +149,7 @@ class VectorServer:
         self._feature_to_handle_if_sql: Optional[Set[str]] = None
         self._valid_serving_keys: Set[str] = set()
         self._serving_initialized: bool = False
+        self.__all_features_on_demand: Optional[bool] = None
 
     def init_serving(
         self,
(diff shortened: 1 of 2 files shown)


python/hsfs/core/vector_server.py
```diff
@@ -415,14 +416,22 @@ def get_feature_vectors(
             request_parameters is None
             or len(request_parameters) == 0
             or isinstance(request_parameters, dict)
-            or len(request_parameters) == len(entries)
-        ), "Request Parameters should be a Dictionary, None, empty or have the same length as the entries"
+            or (not entries or (entries and len(request_parameters) == len(entries)))
```
bubriks, line 419: i don't think `entries and` is necassary since `not entries` covers that case.
```suggestion
            or (not entries or (len(request_parameters) == len(entries)))
```

python/hsfs/core/vector_server.py
```diff
@@ -415,14 +416,22 @@ def get_feature_vectors(
...
+        ), "Request Parameters should be a Dictionary, None, empty or have the same length as the entries when entires is not None or empty."
```
bubriks, line 420:
```suggestion
        ), "Request Parameters should be a Dictionary, None, empty or have the same length as the entries if they are not None or empty."
```

python/hsfs/core/vector_server.py
```diff
@@ -1584,3 +1608,12 @@ def transformed_feature_vector_col_name(self):
             ]
             self._transformed_feature_vector_col_name.extend(output_column_names)
         return self._transformed_feature_vector_col_name
+
+    @property
+    def _all_features_on_demand(self) -> bool:
+        """True if all features in the feature view is on-demand."""
```
bubriks, line 1614:
```suggestion
        """True if all features in the feature view are on-demand."""
```

bubriks (approved):

PR #406, opened 2024-11-20
[FSTORE-1436] Add foreign_keys and make helper columns consistent in materialized training dataset

davitbzh:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/feature_group.py
```diff
@@ -309,24 +311,45 @@ def select_all(
...
         """
-        if include_event_time and include_primary_key:
+        if include_event_time and include_primary_key and include_foreign_key:
             return query.Query(
                 left_feature_group=self,
                 left_features=self._features,
                 feature_store_name=self._feature_store_name,
                 feature_store_id=self._feature_store_id,
             )
         elif include_event_time:
```
manu-sj, line 330: I think this logic would not work now since we have four possible options. e.g. If the users set `include_event_time` = `False`, `include_primary_key` = `True`, `include_foreign_key` = `True` and `include_partition_key` = `False`. The query would not contain foreign keys.

So maybe a better option would be to use a list and append the required keys to it.

"""
removed_keys = []

if not include_event_time:
    removed_keys += [self.event_time]
if not include_primary_key:
    removed_keys += self.primary_key
if not include_foreign_key:
    removed_keys += self.foreign_key
if not include_partition_key:
    removed_keys += self.partition_key

return self.select_except(removed_keys)
"""

python/hsfs/feature_group.py
```diff
@@ -425,7 +448,9 @@ def select_features(
         # Returns
             `Query`. A query object with all features of the feature group.
         """
-        query = self.select_except(self.primary_key + [self.event_time])
+        query = self.select_except(
```
manu-sj, line 451: Function documentation needs to be updated to mention that the function does not select `foreign_keys` also.

!!! info
        This method does not select the primary key, foreign key and event time of the feature group.
        Use `select_all` to include them.
        Note that primary keys do not need to be included in the query to allow joining
        on them.

python/hsfs/feature_view.py
```diff
@@ -1502,9 +1488,6 @@ def create_train_test_split(
         statistics_config: Optional[Union[StatisticsConfig, bool, dict]] = None,
         write_options: Optional[Dict[Any, Any]] = None,
         spine: Optional[SplineDataFrameTypes] = None,
-        primary_key: bool = False,
```
manu-sj, line 1505: I think removing the arguments would be a breaking change, so I am not really sure if we can do it. Would it be better to change the default value of the arguments from `False` to `True`.

python/tests/test_feature_group.py
```diff
@@ -342,8 +344,8 @@ def test_constructor_with_list_event_time_for_compatibility(
     def test_select_all(self):
         query = test_feature_group.select_all()
         features = query.features
-        assert len(features) == 4
-        assert set([f.name for f in features]) == {"pk", "ts", "f1", "f2"}
+        assert len(features) == 5
+        assert set([f.name for f in features]) == {"pk", "fk", "ts", "f1", "f2"}
 
     def test_select_all_exclude_pk(self):
```
manu-sj, line 350: Shouldn't test `test_select_all_exclude_pk` and `test_select_all_exclude_ts` have "fk" in the assert statement.

python/tests/test_feature_group.py
```diff
@@ -46,11 +46,13 @@
     time_travel_format="HUDI",
     partition_key=[],
     primary_key=["pk"],
+    foreign_key=["fk"],
     hudi_precombine_key="pk",
     featurestore_id=1,
     featurestore_name="fs",
     features=[
         feature.Feature("pk", primary=True),
+        feature.Feature("fk", foreign=False),
```
manu-sj, line 55:
```suggestion
        feature.Feature("fk", foreign=True),
```

manu-sj (approved):

PR #407, opened 2024-11-20
[FSTORE-1615] Inserting into streaming FG fails after cluster upgrade

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/engine/python.py
+++ b/python/hsfs/engine/python.py
@@ -1510,7 +1510,7 @@ class Engine:
                 topic_name=feature_group._online_topic_name,
                 feature_store_id=feature_group.feature_store_id,
                 offline_write_options=offline_write_options,
-                high=True,
+                high=False,
             )
             now = datetime.now(timezone.utc)
             feature_group.materialization_job.run(


python/hsfs/engine/python.py
```diff
@@ -1510,7 +1510,7 @@ def _write_dataframe_kafka(
                 topic_name=feature_group._online_topic_name,
                 feature_store_id=feature_group.feature_store_id,
                 offline_write_options=offline_write_options,
-                high=True,
+                high=False,
```
robzor92, line 1513: Not knowing this code, what does this change do?

robzor92 (changes requested):

python/hsfs/engine/python.py
bubriks: `kafka_get_offsets` get offsets for specific topic, the results would look something like this: `topic_name,0:1,1:5,2:6,3:3,4:10`.
Setting `high` to true makes sure we get latest/highest offsets, while setting it to false we get earliest/smallest offsets. Previously it was correct, but it looks like the value was unintentionally changed here: https://github.com/logicalclocks/hopsworks-api/commit/834d13ff1c0b21c6b852ea4fc60397901124cd01#diff-80e02e05a9d4246ee3bdd7a91b698057815fa04a2234772c883aba8b8770f977L1402

PR #428, opened 2024-12-09
[FSTORE-1630] Model Dependent Transformation Functions creates feature names that are longer than 64 character causing logging feature group ingestion to fail

manu-sj:
This PR adds an `alias` function to transformation function that can be used to specify output column names for the transformation functions.

JIRA Issue: https://hopsworks.atlassian.net/browse/FSTORE-1630

Priority for Review: -

Related PRs: 
https://github.com/logicalclocks/hopsworks-ee/pull/2168
https://github.com/logicalclocks/loadtest/pull/493
https://github.com/logicalclocks/logicalclocks.github.io/pull/429
https://github.com/logicalclocks/hopsworks-front/pull/1403

**How Has This Been Tested?**

- [x] Unit Tests
- [x] Integration Tests
- [x] Manual Tests on VM

```
```


python/hsfs/hopsworks_udf.py
```diff
@@ -173,6 +176,8 @@ def __init__(
         dropped_argument_names: Optional[List[str]] = None,
         dropped_feature_names: Optional[List[str]] = None,
         feature_name_prefix: Optional[str] = None,
+        output_column_names: Optional[str] = None,
+        generate_output_col_names: bool = True,
```
bubriks, line 180: doesn't match the description above

python/hsfs/transformation_function.py
```diff
@@ -78,12 +83,17 @@ def __init__(
                 "Please use the hopsworks_udf decorator when defining transformation functions."
             )
 
-        self._hopsworks_udf: HopsworksUdf = hopsworks_udf
+        self.__hopsworks_udf: HopsworksUdf = hopsworks_udf
```
bubriks, line 86: why change this?

python/tests/test_transformation_function.py
```diff
@@ -454,3 +461,346 @@ def test_func(col1, statistics=stats):
...
+            transformation_type=TransformationType.ON_DEMAND,
+        )
+
+        with pytest.raises(FeatureStoreException) as exp:
+            odt.alias({"name": "col1"})
+
+        assert (
+            str(exp.value)
+            == "Invalid output feature names provided for the transformation function 'add_one(feature)'. Please ensure all arguments are strings."
+        )
```
bubriks, line 619: i think test should self enclosed, here we have 2 asserts for actions on two different TransformationFunctions, we could just split it up to make sure that if one/both fail its faster to identify the issue. (applies in multiple cases)

Maybe just to avoid duplication you can parametrize: https://docs.pytest.org/en/stable/how-to/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions

bubriks (changes requested):

python/hsfs/hopsworks_udf.py
manu-sj: Corrected description.

python/tests/test_transformation_function.py
manu-sj: Agreed, I splited the test for now since that was a quick solution.

I think parametrizing the test would be the best solution but that could take some work because corresponding fixtures would also have to be created and I would have to think about how to model the fixtures for the test so that it is easy to add testcase. So would it be okey if I create another JIRA for this and handle it there.

bubriks (approved):

PR #440, opened 2024-12-16
[FSTORE-1645] Catch HSML initialization permission-related errors in login

aversey:
JIRA Issue: [FSTORE-1645](https://hopsworks.atlassian.net/browse/FSTORE-1645)

[FSTORE-1645]: https://hopsworks.atlassian.net/browse/FSTORE-1645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/hopsworks_common/connection.py
+++ b/python/hopsworks_common/connection.py
@@ -440,7 +440,7 @@ class Connection:
             except RestAPIError as e:
                 if e.response.error_code == 403 and e.error_code == 320004:
                     print(
-                        'API key does not include "SERVING" scope, the related functionality will be disabled.'
+                        'The used API key does not include "SERVING" scope, the related functionality will be disabled.'
                     )
                     _logger.debug(f"The ignored exception: {e}")
                 else:
--- a/python/hopsworks_common/connection.py
+++ b/python/hopsworks_common/connection.py
@@ -25,6 +25,7 @@ import weakref
 from typing import Any, Optional
 
 from hopsworks_common import client, constants, usage, util, version
+from hopsworks_common.client.exceptions import RestAPIError
 from hopsworks_common.core import (
     hosts_api,
     project_api,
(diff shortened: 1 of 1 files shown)


python/hopsworks_common/connection.py
```diff
@@ -429,7 +430,17 @@ def _provide_project(self, name=None):
...
             # load_default_configuration has to be called before using hsml
             # but after a project is provided to client
-            self._model_serving_api.load_default_configuration()  # istio client, default resources,...
+            try:
+                # istio client, default resources,...
+                self._model_serving_api.load_default_configuration()
+            except RestAPIError as e:
+                if e.response.error_code == 403:
+                    print(
+                        "You lack permission to initialize model serving. Machine learning functionality will be not available."
```
vatj, line 439:
```suggestion
                       "Api key does not include "SERVING" scope, related functionality will be disabled."
```

python/hopsworks_common/connection.py
```diff
@@ -429,7 +430,17 @@ def _provide_project(self, name=None):
...
+                    )
+                    print("The permission-related exception is:", e)
```
vatj, line 441: Can you use a debug logger for that instead?

vatj (changes requested):

python/hopsworks_common/connection.py
aversey: This message may be misleading, since the code catches all 403. I'll think about this and hopefully come up with a better solution.

python/hopsworks_common/connection.py
aversey: I printed it for the same reason as is explained above. I will change it to use the debug logger.

vatj (approved):

PR #455, opened 2025-01-23
[FSTORE-1523] Python API should return None if get function could not find backend entity

robzor92:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hopsworks_common/core/environment_api.py
+++ b/python/hopsworks_common/core/environment_api.py
@@ -17,7 +17,7 @@
 import json
 from typing import List, Optional
 
-from hopsworks_common import client, environment, usage
+from hopsworks_common import client, decorators, environment, usage
 from hopsworks_common.engine import environment_engine
 
 
@@ -54,7 +54,7 @@ class EnvironmentApi:
         # Returns
             `Environment`: The Environment object
         # Raises
-            `RestAPIError`: If unable to create the environment
+            `hopsworks.client.exceptions.RestAPIError`: If the backend encounters an error when handling the request
         """
         _client = client.get_instance()
 
@@ -83,8 +83,23 @@ class EnvironmentApi:
 
     @usage.method_logger
     def get_environments(self) -> List[environment.Environment]:
-        """
-        Get all available python environments in the project
+        """Get all available environments in the project.
+
+        ```python
+
+        import hopsworks
+
+        project = hopsworks.login()
+
+        env_api = project.get_environment_api()
+
+        envs = env_api.get_environments()
+
+        ```
+        # Returns
+            `List[Environment]`: List of Environment objects
+        # Raises
+            `hopsworks.client.exceptions.RestAPIError`: If the backend encounters an error when handling the request
         """
         _client = client.get_instance()
 
@@ -98,8 +113,11 @@ class EnvironmentApi:
         )
 
     @usage.method_logger
-    def get_environment(self, name: str) -> environment.Environment:
-        """Get handle for the Python environment for the project
+    @decorators.catch_not_found(
+        ["hopsworks_common.environment.Environment"], fallback_return=None
+    )
+    def get_environment(self, name: str) -> Optional[environment.Environment]:
+        """Get handle for a Python environment in the project
 
         ```python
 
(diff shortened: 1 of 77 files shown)


python/hopsworks_common/core/environment_api.py
```diff
@@ -99,7 +99,7 @@ def get_environments(self) -> List[environment.Environment]:
 
     @usage.method_logger
     def get_environment(self, name: str) -> environment.Environment:
```
aversey, line 101: I think the type should be Optional[environment.Environment], the same in the docstring.

python/hopsworks_common/core/environment_api.py
```diff
@@ -115,20 +115,26 @@ def get_environment(self, name: str) -> environment.Environment:
...
-                "GET", path_params, query_params=query_params, headers=headers
+        try:
+            return environment.Environment.from_response_json(
+                _client._send_request(
+                    "GET", path_params, query_params=query_params, headers=headers
+                )
             )
-        )
+        except RestAPIError as e:
+            if e.response.json().get("errorCode", "") == 300000 and e.response.status_code == 404:
```
aversey, line 134: I think a cleaner solution would be to add a decorator like this:
```
def catch_not_found(object_class):
    def decorator(f):
        @functools.wraps(f)
        def g(*args, **kwds):
            try:
                return f(*args, **kwds)
            except RestAPIError as e:
                if e.response.status_code == 404 and e.response.json().get("errorCode", "") == object_class.NOT_FOUND_ERROR_CODE:
                    return None
                else:
                    raise e
        return g
    return decorator
```
And then just add to each "object" class its "not found" error code and wrap the getters in this decorator, like this:
```
@usage.method_logger
@catch_not_found(environment.Environment)
def get_environment(self, name: str) -> Optional[environment.Environment]: ...
```
And:
```
class Environmenet:
    NOT_FOUND_ERROR_CODE = 300000
```
And so on.

robzor92: Great feedback, will try to implement this where it is possible which should be in most places

python/hopsworks_common/decorators.py
```diff
@@ -28,6 +28,8 @@
     polars_not_installed_message,
 )
 
+from hopsworks.client.exceptions import RestAPIError
```
aversey, line 31: Move the import line into catch_not_found to solve the circimports issue. It should be totally fine in terms of performance, since decorators are called only once.

python/hopsworks_common/decorators.py
```diff
@@ -49,6 +50,36 @@ def if_connected(inst, *args, **kwargs):
     return if_connected
 
 
+def catch_not_found(class_import_path_list, fallback_return=None):
```
aversey, line 53: I think that something like:
```
def catch_not_found(*class_objects, fallback_return=None):
```
Maybe easier to use (linters would be able to check that the object is imported correctly, plus there will be no need to wrap them in a list). Is there a reason why do you dynamically import classes?

python/hopsworks_common/project.py
```diff
@@ -202,10 +202,6 @@ def get_job_api(self):
         """
         return self._job_api
 
-    def get_jobs_api(self):
```
aversey, line 205: Is it ok to remove it now?

python/hopsworks_common/core/tags_api.py
```diff
@@ -95,12 +96,22 @@ def get(self, metadata_instance, name: str = None, training_dataset_version=None
...
-            )
-        }
+        try:
+            return {
+                tag._name: json.loads(tag._value)
+                for tag in tag.Tag.from_response_json(
+                    _client._send_request("GET", path_params)
+                )
+            }
+        # migrate to catch_not_found decorator later on
```
aversey, line 106: It seems that you forgot to do it. =)

aversey (commented):
It is nice!

I think that maybe explicit `return None` is a bit more readable, but I like the conciseness of the implicit one, so it is fine to keep it as it is.

I also have a comment on the decorator.

python/hopsworks_common/decorators.py
robzor92: I had to add the dynamic import because otherwise I kept getting circular import issues, without actually adding new imports. To be honest I don't understand it. I wonder if it is because the decorators are initialized when a module is imported. I can let you know how to reproduce it if you want

robzor92: Adapted the function to `def catch_not_found(*class_objects, fallback_return=None):`

python/hopsworks_common/project.py
robzor92: Based on the feedback I got I excluded the function from the docs, but keep it in the API for now

aversey (approved):

PR #466, opened 2025-02-03
[FSTORE-1682] Change default behaviour of get_feature_vector to return on-demand features by default

manu-sj:
This PR introduces a **breaking change**.

It changes the behaviour of `get_feature_vector` so as a disconnect, on-demand features from the `transform` argument. 

In the updated behaviour when:
- `transform=True` and `on_demand_feature=True/False` -  Return the complete transfomed feature vector with on-demand features and model-dependent features.
- `transform=False` and `on_demand_feature=True` - Return feature vector with on-demand features but no model-dependent transformations applied.
- `transform=False` and `on_demand_feature=False` - Return untransformed feature vector without on-demand features or model-dependent transformations applied.

JIRA Issue: https://hopsworks.atlassian.net/browse/FSTORE-1682

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [x] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/vector_server.py
+++ b/python/hsfs/core/vector_server.py
@@ -372,6 +372,7 @@ class VectorServer:
         force_rest_client: bool = False,
         force_sql_client: bool = False,
         transform: bool = True,
+        on_demand_features: Optional[bool] = True,
         request_parameters: Optional[Dict[str, Any]] = None,
         transformation_context: Dict[str, Any] = None,
     ) -> Union[pd.DataFrame, pl.DataFrame, np.ndarray, List[Any], Dict[str, Any]]:
@@ -412,6 +413,7 @@ class VectorServer:
             allow_missing=allow_missing,
             client=online_client_choice,
             transform=transform,
+            on_demand_features=on_demand_features,
             request_parameters=request_parameters,
             transformation_context=transformation_context,
         )
@@ -422,7 +424,7 @@ class VectorServer:
             inference_helper=False,
             return_type=return_type,
             transform=transform,
-            on_demand_feature=transform,
+            on_demand_feature=on_demand_features,
         )
 
     def get_feature_vectors(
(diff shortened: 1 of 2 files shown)


python/hsfs/core/vector_server.py
```diff
@@ -371,7 +371,8 @@ def get_feature_vector(
         allow_missing: bool = False,
         force_rest_client: bool = False,
         force_sql_client: bool = False,
-        transform: bool = True,
+        transform: bool = False,
```
SirOibaf, line 374: I think this in an unecessary brekaing change to be honest.

python/hsfs/feature_view.py
```diff
@@ -607,7 +608,8 @@ def get_feature_vector(
             force_sql_client: boolean, defaults to False. If set to True, reads from online feature store
                 using the SQL client if initialised.
             allow_missing: Setting to `True` returns feature vectors with missing values.
-            transformed: Setting to `False` returns the untransformed feature vectors.
+            transform: `bool`, defaults to `True`. Setting this to `False` returns a feature vector that includes both precomputed and on-demand features without applying model-dependent transformations. When set to `True`, it automatically sets `on_demand_features` to `True` and applies model-dependent transformations.
```
SirOibaf, line 611: I think the documentation here is a bit off?

python/hsfs/core/vector_server.py
manu-sj: The change in the default behavior in transform was from Jim's recommendation. 

His reasoning was that usually everyone would want to log untransformed and transformed features. So changing the default behaviour allows for the code to be little bit more cleaner. 

```
untransformed_feature_vector = fv.get_feature_vector(entry={"sk":1}) 

transformed_feature_vector = fv.transform(feature_vector)

predictions = model.predict(transformed_feature_vector)

fv.log(untransformed_feature_vector, transformed_feature_vector, predictions)
```

Changing the default behavior for transformation functions like you mentioned has a much higher impact (the code would still work but we get untransformed features instead harder to debug, might go unnoticed by users) than disconnecting the application of ODT and MDT. 

So I am okey to revert this change.

SirOibaf (approved):

PR #512, opened 2025-03-18
[FSTORE-1714] Query Construction Problem

manu-sj:
This PR does the following changes:
1. Uses a `fully qualified name` to materialize and read `primary key` and `event time` in training data when not explicitly selected by the user.
2. A fully qualified name is obtained by containing the feature store name, feature group name and version with the the feature name. `feature_store_name (project_name) + feature_group_name_version + feature name`
3. Change automatically generated prefixes to use the feature groups name instead of aliased feature group.
4. Added log messages when prefixes are automatically added.
5. Handled extraction of primary keys in the API for all nested joins in the case of snowflake schema

JIRA Issue: https://hopsworks.atlassian.net/browse/FSTORE-1714

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/core/feature_view_engine.py
```diff
@@ -1070,52 +1072,159 @@ def _get_feature_view_url(self, fv: "feature_view.FeatureView"):
         )
         return util.get_hostname_replaced_url(path)
 
-    def _get_primary_keys_from_query(self, fv_query_obj, check_duplicate=True):
-        fv_pks = set(
-            [
+    def generate_fully_qualied_feature_name(
```
bubriks, line 1075: why generate fully_qualied_feature_name both in api and backend? maybe better to limit it to one place?

python/hsfs/core/feature_view_engine.py
```diff
@@ -1070,52 +1072,159 @@ def _get_feature_view_url(self, fv: "feature_view.FeatureView"):
...
+        return list(pk_names)
+
+    def _get_eventtimes_from_query(
+        self, query: Query, check_duplicate: bool = True
+    ) -> List[str]:
+        """
+        Function that checks the event times from the query object and returns them as a list.
+
+        #Arguments:
+            fv_query_obj : `Query`. Query object from which the event times are extracted.
```
bubriks, line 1203: doesnt match the actual args (other places as well)

python/hsfs/core/feature_view_engine.py
bubriks: also why do it in feature view engine?

python/hsfs/core/feature_view_engine.py
manu-sj: Correct arg names in comments

python/hsfs/core/feature_view_engine.py
manu-sj: I moved the function to the util file.

I think you are right,  the function `fully_qualied_feature_name` should only be present in the backend. But I think there would be quite some changes required to do this.

The problem is the the primary key and event time names generated from this function is used to [drop the `event_time` and `primary_key` while reading an already materialized training data](https://github.com/logicalclocks/hopsworks-api/blob/main/python/hsfs/core/feature_view_engine.py#L693), since we are always materializing the primary keys and event time. 

We use the fully qualified name when user does not manually select the `primary_key` and `event_time` in the query. So in this case the client does not get any information regarding the names to be used since the feature themself are in the query DTO. So that is why I have manually created the names here.

One idea is to update the feature's in the feature group DTO in the query to have fully qualified names if it is required. But I think it might make the code much more complex, or break something else. So would it okey to have this as a separate PR.

bubriks (approved):

SirOibaf (approved):

PR #521, opened 2025-03-28
[FSTORE-1698] Data sources

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


aversey (commented):
It seems that it is not backwards compatible (in the sense that this version of Python API cannot be used against an older cluster). It is better to double-check if it is okay.

python/hsfs/core/arrow_flight_client.py
```diff
@@ -646,6 +646,79 @@ def _serialize_featuregroup_connector(fg, query, on_demand_fg_aliases):
         connector["time_travel_type"] = "hudi"
     return connector
 
+def _get_connector_options(fg):
```
manu-sj, line 649: Want do you think about moving this function to the `StorageConnector` class. We could define something like `get_connector_options(data_source)` and since we already have different class for each storage type we could have different implementations of the same function.

python/hsfs/core/arrow_flight_client.py
```diff
@@ -646,6 +646,79 @@ def _serialize_featuregroup_connector(fg, query, on_demand_fg_aliases):
...
+    elif connector_type == StorageConnector.S3:
+        option_map = {
+            "access_key": connector.access_key,
+            "secret_key": connector.secret_key,
+            "session_token": connector.session_token,
+            "region": connector.region,
+        }
+        if connector.arguments.get("fs.s3a.endpoint"):
+            option_map["endpoint"] = connector.arguments.get("fs.s3a.endpoint")
+        option_map["path"] = fg.location
```
manu-sj, line 715: Should this be `datasource.path` instead of `fg.location`

manu-sj: From looking through the code a little bit I dont think fg._location is not set from anywhere in the python client or in the backend. And it is only used through the function [prepare_spark_location](https://github.com/logicalclocks/hopsworks-api/blob/main/python/hsfs/feature_group.py#L2115) which basically returns the storage connector path.

Maybe I am missing something here.

python/hsfs/engine/spark.py
```diff
@@ -194,11 +194,11 @@ def set_job_group(self, group_id, description):
     def register_external_temporary_table(self, external_fg, alias):
         if not isinstance(external_fg, fg_mod.SpineGroup):
             external_dataset = external_fg.storage_connector.read(
-                external_fg.query,
+                external_fg._data_source.query,
```
manu-sj, line 197:
```suggestion
                external_fg.data_source.query,
```

python/hsfs/engine/spark.py
```diff
@@ -194,11 +194,11 @@ def set_job_group(self, group_id, description):
...
                 external_fg.data_format,
                 external_fg.options,
                 external_fg.storage_connector._get_path(
-                    external_fg.path
+                    external_fg._data_source.path
```
manu-sj, line 201:
```suggestion
                    external_fg.data_source.path
```

python/hsfs/feature_group.py
```diff
@@ -141,8 +142,13 @@ def __init__(
...
             ]
         ] = None,
+        data_source: Optional[
+            Union[
+                ds.DataSource,
+                Dict[str, Any],
+            ]
+        ] = None,
         storage_connector: Union[sc.StorageConnector, Dict[str, Any]] = None,
-        path: Optional[str] = None,
```
manu-sj, line 145: Wouldn't this be a breaking change? Maybe should we keep the `path` variable for now and set `data_source.path` internally and and throw a `DeprecationWarning` for this usecase.

python/hsfs/feature_store.py
```diff
@@ -1108,9 +1108,7 @@ def create_external_feature_group(
         """
         feature_group_object = feature_group.ExternalFeatureGroup(
             name=name,
-            query=query,
```
manu-sj, line 1111: Wouldn't these changes be breaking. I think maybe we can add the new parameter `data_sources` and still keep `path` and `query` for now, and set them internally. We could also throw a `DeprecationWarning` in case the user provides `query` or `path`.

Same is true for all the feature group creation functions.

python/tests/test_feature_group.py
```diff
@@ -694,9 +694,9 @@ def test_from_response_json(self, backend_fixtures):
 
         # Assert
         assert isinstance(fg.storage_connector, storage_connector.StorageConnector)
-        assert fg.query == "Select * from "
+        assert fg._data_source.query == "Select * from "
```
manu-sj, line 697:
```suggestion
        assert fg.data_source.query == "Select * from "
```

python/tests/test_feature_group.py
```diff
@@ -694,9 +694,9 @@ def test_from_response_json(self, backend_fixtures):
...
         assert fg.data_format == "HUDI"
-        assert fg.path == "test_path"
+        assert fg._data_source.path == "test_path"
```
manu-sj, line 699:
```suggestion
        assert fg.data_source.path == "test_path"
```

python/tests/test_feature_group.py
```diff
@@ -728,9 +728,9 @@ def test_from_response_json_list(self, backend_fixtures):
         assert len(fg_list) == 1
         fg = fg_list[0]
         assert isinstance(fg.storage_connector, storage_connector.StorageConnector)
-        assert fg.query == "Select * from "
+        assert fg._data_source.query == "Select * from "
```
manu-sj, line 731:
```suggestion
        assert fg.data_source.query == "Select * from "
```

python/tests/test_feature_group.py
```diff
@@ -728,9 +728,9 @@ def test_from_response_json_list(self, backend_fixtures):
...
         assert fg.data_format == "HUDI"
-        assert fg.path == "test_path"
+        assert fg._data_source.path == "test_path"
```
manu-sj, line 733:
```suggestion
        assert fg.data_source.path == "test_path"
```

python/tests/test_feature_group.py
```diff
@@ -760,9 +760,9 @@ def test_from_response_json_basic_info(self, backend_fixtures):
 
         # Assert
         assert isinstance(fg.storage_connector, storage_connector.StorageConnector)
-        assert fg.query is None
+        assert fg._data_source.query is None
```
manu-sj, line 763:
```suggestion
        assert fg.data_source.query is None
```

python/tests/test_feature_group.py
```diff
@@ -760,9 +760,9 @@ def test_from_response_json_basic_info(self, backend_fixtures):
...
         assert fg.data_format is None
-        assert fg.path is None
+        assert fg._data_source.path is None
```
manu-sj, line 765:
```suggestion
        assert fg.data_source.path is None
```

python/hsfs/core/arrow_flight_client.py
bubriks: no `location` should remain, i know naming is confusing but `location` uses `path` and is generated in the backend.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/SparkEngine.java
```diff
@@ -214,10 +215,14 @@ public Dataset<Row> sql(String query) {
 
   public Dataset<Row> registerOnDemandTemporaryTable(ExternalFeatureGroup onDemandFeatureGroup, String alias)
       throws FeatureStoreException, IOException {
+    DataSource dataSource = onDemandFeatureGroup.getDataSource();
+    dataSource.setPath(onDemandFeatureGroup.getStorageConnector().getPath(
+        onDemandFeatureGroup.getDataSource().getQuery()));
```
manu-sj, line 220: Should this be `onDemandFeatureGroup.getDataSource().getPath()` instead of `getQuery()` because I see that we had `getStorageConnector().getPath(onDemandFeatureGroup.getPath())` earlier

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/TrainingDatasetEngine.java
```diff
@@ -96,8 +97,10 @@ public Dataset<Row> read(TrainingDataset trainingDataset, String split, Map<Stri
     } else {
       path = new Path(trainingDataset.getLocation(), trainingDataset.getName()).toString();
     }
-    return storageConnectorUtils.read(trainingDataset.getStorageConnector(), null,
-        trainingDataset.getDataFormat().toString(), readOptions, path);
+    DataSource dataSource = new DataSource();
+    dataSource.setQuery(path);
```
manu-sj, line 101: Should this be `setPath`

python/hsfs/core/arrow_flight_client.py
manu-sj: Aah yes I got it location is in the base class of FeaturegroupDTO

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/SparkEngine.java
bubriks: good catch

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/TrainingDatasetEngine.java
bubriks: good catch x2

python/hsfs/core/data_source_api.py
```diff
@@ -0,0 +1,94 @@
...
+            "tables",
+        ]
+
+        query_params = {"database": database}
+
+        return ds.DataSource.from_response_json(
+            _client._send_request("GET", path_params, query_params)
+        )
+
+    def get_data(self, feature_store_id: int, name: str, data_source: ds.DataSource) -> dsd.DataSourceData:
```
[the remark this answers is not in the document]
bubriks: yeah i think for now we can leave it as it is, it is intended only for LLM assistant to use in V1.

manu-sj (approved):

PR #531, opened 2025-04-10
[FSTORE-1724] Event time columns not casted to lower case before feature name validation

manu-sj:
This PR fixes an issue in which event time is not automatically casted to lower case to be compatible with feature store schema.

JIRA Issue: https://hopsworks.atlassian.net/browse/FSTORE-1724

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hopsworks_common/util.py
```diff
@@ -121,8 +122,24 @@ def validate_embedding_feature_type(embedding_index, schema):
             )
 
 
-def autofix_feature_name(name: str) -> str:
+def autofix_feature_name(name: str, warn: bool = False) -> str:
```
SirOibaf, line 125: I have several questions here: 
- We already have logging for both capital case letters and underscore here - https://github.com/logicalclocks/hopsworks-api/blob/93c5d607b3aeb21268f2443a9ac84ac08cc4448f/python/hsfs/engine/python.py#L719 (and in the Spark engine as well) - Can we avoid the redundancy?
- I see that you have added a `warn` flag, but it's always set to `True` when the method is called. What am I missing?

SirOibaf (changes requested):

python/hopsworks_common/util.py
manu-sj: Question 1:
- Regarding redundancy- The log messages are slightly different. 
    In the python/spark engines the log message mention that the _inserted dataFrame_ containing upper case letters / spaces. In my case the log message mentions that the _feature_ contains upper case letters or spaces. Since the messages themselves are it might be make the code compl I moved the checks to the `utils` file so that the logic used to print the warning message is not duplicated.

Question 2:
- `autofix_feature_name` is used in many places eg:
 https://github.com/logicalclocks/hopsworks-api/blob/609d0ddfd8296d25cbed529b0fd72addf0c994ee/python/hsfs/training_dataset_feature.py#L41
https://github.com/logicalclocks/hopsworks-api/blob/609d0ddfd8296d25cbed529b0fd72addf0c994ee/python/hsfs/core/feature_descriptive_statistics.py#L71
https://github.com/logicalclocks/hopsworks-api/blob/609d0ddfd8296d25cbed529b0fd72addf0c994ee/python/hsfs/feature_view.py#L4041

    I did not want to show warning for all these case. I only wanted to show warnings when users [defined a feature](https://github.com/logicalclocks/hopsworks-api/blob/609d0ddfd8296d25cbed529b0fd72addf0c994ee/python/hsfs/feature.py#L66) for a feature group or if [any of the feature names](https://github.com/logicalclocks/hopsworks-api/pull/531/files#diff-4ac064b1d72d926de7a5387b7d15fa9b889ac5a05ba8bcbc9974d9de334a5435R2390) specified while creating a feature group have upper case / spaces. 

    Since the `autofix_feature_name` is used widely to sanitize feature names I though the it would be better to add a `warn` parameter and then set it to `True` whenever required. 
    
    The alternative was to add a new function that does this check and call it wherever required. Do you think that would be a better alternative?

SirOibaf: Ok sorry - I missed the additional usages of the `autofix_feature_name`

Regarding 1 - I see that the call in the python/spark engines don't trigger a warning, so we shouldn't have a double warning message.

SirOibaf (approved):

python/hopsworks_common/util.py
manu-sj: Aah sorry I missed your comment. No we would not have double warning by default python suppresses duplicate warning based on source code location, warning text and warning location.

https://docs.python.org/3/library/warnings.html
> The disposition of warnings can vary based on the [warning category](https://docs.python.org/3/library/warnings.html#warning-categories), the text of the warning message, and the source location where it is issued. Repetitions of a particular warning for the same source location are typically suppressed.

PR #536, opened 2025-04-17
[FSTORE-1617] Integrate delta-rs HopsFS writer to hsfs python client

davitbzh:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/feature_store.py
```diff
@@ -500,7 +500,7 @@ def create_feature_group(
         version: Optional[int] = None,
         description: Optional[str] = "",
         online_enabled: Optional[bool] = False,
-        time_travel_format: Optional[str] = "HUDI",
+        time_travel_format: Optional[str] = None,
```
bubriks, line 503: does it still default to Hudi? just in the back end?

python/hsfs/feature_group.py
```diff
@@ -2399,8 +2398,15 @@ def __init__(
         else:
             # initialized by user
             # for python engine we always use stream feature group
-            if engine.get_type() == "python":
-                self._stream = True
+            if time_travel_format is None:
+                if engine.get_type() == "python":
+                    if not online_enabled and not self._stream:
+                        self._time_travel_format = "DELTA"
```
bubriks, line 2404: not sure i follow these changes.

python/hsfs/core/delta_engine.py
```diff
@@ -168,6 +278,51 @@ def _generate_merge_query(self, source_alias, updates_alias):
         megrge_query_str = " AND ".join(merge_query_list)
         return megrge_query_str
 
+    @staticmethod
+    def _get_last_commit_metadata_delta_rs(base_path):
```
bubriks, line 282: maybe we could change it so that there would be less duplication with `def _get_last_commit_metadata(`?

python/hsfs/core/delta_engine.py
```diff
@@ -85,15 +98,23 @@ def _setup_delta_read_opts(self, delta_fg_alias, read_options):
...
+            location = self._feature_group.location.replace("hopsfs", "hdfs")
+            try:
+                fg_source_table = DeltaRsTable(location)
+                is_delta_table = True
+            except TableNotFoundError:
+                is_delta_table = False
+
+        if not is_delta_table:
             raise FeatureStoreException(
                 f"This is no data available in Feature group {self._feature_group.name}, or it not DELTA enabled "
```
bubriks, line 115: there is no data available in Feature group

bubriks (changes requested):

python/hsfs/feature_store.py
davitbzh: Python defaults to `stream=True` because it was pain to write to hopsfs. In legacy FG we were uploading dataframe via rest api and then spark was writing it. Once we introduced stream enabled feature group it was much easier and convinient to write to Kafka and then start deltastreamer. But here problem is that launching spark jobs takes lots of time to write (even couple of thousands of rows).  

To address above delta-rs can directly write into offline table (parquet) files. But this comes with the problem that how can we populate online tables? We don't have flexibility for this. So the compromise is that if you want to have fast writes using python client and you don't require online tables then you can use delta-rs. This also address issues we have with serverless where most of the feature groups are offline

python/hsfs/feature_group.py
davitbzh: comprehensive answer above :)

python/hsfs/engine/python.py
```diff
@@ -815,6 +817,18 @@ def save_dataframe(
             return self._write_dataframe_kafka(
                 feature_group, dataframe, offline_write_options
             )
+        elif engine.get_type() == "python":
+            if (
+                not online_enabled and not feature_group.stream
```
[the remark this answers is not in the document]
davitbzh: comprehensive answer above :)

python/hsfs/core/delta_engine.py
davitbzh: This needs more testing but to me logfiles seem different. for example in spark Delta there is camel case names but in detla-rs snake case names: `num_target_rows_inserted`

PR #544, opened 2025-05-07
[FSTORE-1732] Handle duplicated events for Delta Table materializations jobs

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/engine/spark.py
+++ b/python/hsfs/engine/spark.py
@@ -650,24 +650,36 @@ class Engine:
         feature_group: Union[fg_mod.FeatureGroup, fg_mod.ExternalFeatureGroup],
         dataframe: Union[RDD, DataFrame],
     ):
+        data_column = "value"
         """
-        Deserializes 'value' column from binary using avro schema.
+        Step 1: Deserializes 'value' column from binary using avro schema.
         """
         decoded_dataframe = dataframe.withColumn(
-            "value", from_avro("value", feature_group._get_encoded_avro_schema())
+            data_column, from_avro(data_column, feature_group._get_encoded_avro_schema())
         )
 
-        """Decodes all complex type features from binary using their avro type as schema."""
-        return decoded_dataframe.select(
-            [
-                from_avro(
-                    field_name, feature_group._get_feature_avro_schema(field_name.removeprefix("value."))
+        """
+        Step 2: Replace complex fields in the 'value' column
+        """
+        new_value_fields = []
+        for field in json.loads(feature_group.avro_schema)["fields"]:
+            field_name = field["name"]
+            if field_name in feature_group.get_complex_features():
+                # re-apply from_avro on the nested field
+                decoded_field = from_avro(
+                    col(f"{data_column}.{field_name}"),
+                    feature_group._get_feature_avro_schema(field_name)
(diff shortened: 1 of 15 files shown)


python/hsfs/engine/spark.py
```diff
@@ -651,23 +651,25 @@ def _deserialize_from_avro(
...
+        for field in json.loads(feature_group.avro_schema)["fields"]:
+            name = field["name"]
+            full_field = f"value.{name}"
+
+            if name in feature_group.get_complex_features():
+                decoded_dataframe = decoded_dataframe.withColumn(
+                    full_field, from_avro(col(full_field), feature_group._get_feature_avro_schema(name))
+                )
+            else:
+                decoded_dataframe = decoded_dataframe.withColumn(full_field, col(full_field))
```
SirOibaf, line 670: This is not a good idea. Using `withColumn` within a for loop makes Spark super slow and potentially breaks Spark. 
This is because every `withColumn` adds a new operation in the Spark Tree that the compiler and optimizer need to deal with.

You should assemble the list of columns you need and then do a single Spark operation - (like the code was before).

utils/python/hsfs_utils.py
```diff
@@ -325,9 +326,21 @@ def offline_fg_materialization(spark: SparkSession, job_conf: Dict[Any, Any], in
     # deserialize dataframe so that it can be properly saved
     deserialized_df = engine.get_instance()._deserialize_from_avro(entity, filtered_df)
 
+    # de-duplicate records
+    # timestamp cannot be relied on to order the records in case of duplicates, if they are produced together they would have the same timestamp.
+    # Instead use offset to order the records, they are strictly increasing within a partition and since we use primary keys for generating Kafka message keys duplicates are guaranteed to be in the same partition.
+    window = Window.partitionBy([f"value.{key}" for key in entity.primary_key]) \
```
SirOibaf, line 332: Remember that in the offline store the unique identifier of the row is `primary key` + `event time` so here the `partitionBy` should also include the event time.

utils/python/hsfs_utils.py
```diff
@@ -325,9 +326,21 @@ def offline_fg_materialization(spark: SparkSession, job_conf: Dict[Any, Any], in
...
+    # get only the feture values (remove kafka metadata)
+    deduped_df = deduped_df.select("value.*")
+
     # insert data
     entity.stream = False # to make sure we dont write to kafka
-    entity.insert(deserialized_df, storage="offline")
+    entity.insert(deduped_df, storage="offline")
 
     # update offsets
     df_offsets = (df if limit > filtered_df.count() else filtered_df).groupBy('partition').agg(max('offset').alias('offset')).collect()
```
SirOibaf, line 346: Not specifically related to this PR - but this `filtered_df` is not cached - so the `count()` might recompute the thing from the beginning and might be different from what it was written in the offline store.

Are we missing some PRs here?

SirOibaf (changes requested):

SirOibaf (approved):

PR #558, opened 2025-05-22
[FSTORE-1750] Allow feature vector retrieval from feature view created using offline feature groups with only on-demand features

manu-sj:
This PR add support for retrieving feature vectors from feature view with offline feature group if it contains only on-demand features

JIRA Issue: https://hopsworks.atlassian.net/browse/FSTORE-1750

Priority for Review: -

Related PRs: https://github.com/logicalclocks/loadtest/pull/576

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hopsworks_common/core/job_api.py
+++ b/python/hopsworks_common/core/job_api.py
@@ -183,7 +183,7 @@ class JobApi:
         """
         _client = client.get_instance()
 
-        config = util.validate_job_conf(config, self._project_name)
+        config = util.validate_job_conf(config, _client._project_name)
 
         path_params = ["project", _client._project_id, "jobs", name]
 


python/hsfs/core/vector_server.py
```diff
@@ -194,14 +226,14 @@ def init_serving(
...
-        if self._init_rest_client:
+        if self._init_rest_client and self.__all_feature_groups_online:
             self.setup_rest_client_and_engine(
                 entity=entity,
                 config_rest_client=config_rest_client,
                 reset_rest_client=reset_rest_client,
             )
 
-        if self._init_sql_client:
+        if self._init_sql_client and self.__all_feature_groups_online:
```
kennethmhc, line 236:
```suggestion
        if self._init_sql_client and self._all_parent_feature_groups_online:
```

python/hsfs/core/vector_server.py
```diff
@@ -194,14 +226,14 @@ def init_serving(
         self.init_transformation(entity)
         self.set_return_feature_value_handlers(features=entity.features)
 
-        if self._init_rest_client:
+        if self._init_rest_client and self.__all_feature_groups_online:
```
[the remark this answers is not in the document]
manu-sj: Updated as per comments

kennethmhc (approved):

python/hsfs/core/vector_server.py
```diff
@@ -165,6 +181,20 @@ def init_serving(
         default_client: Optional[Literal["rest", "sql"]] = None,
     ):
         self._training_dataset_version = training_dataset_version
+
+        self._parent_feature_groups = entity.get_parent_feature_groups().accessible
```
SirOibaf, line 185: Any particular reason you chose to use provenance here instead of relying on the query in the feature view? I think we don't usually rely on provenance for this type of information.

manu-sj: The main reason why I did not do it was to remove code complexity. I could do something similar to what we are doing here for [primary_keys](https://github.com/logicalclocks/hopsworks-api/blob/main/python/hsfs/core/feature_view_engine.py#L1154C9-L1154C37) and [event_time](https://github.com/logicalclocks/hopsworks-api/blob/main/python/hsfs/core/feature_view_engine.py#L1187), but it would require a recursive function (because of snowflake schemas). 

So I though that using provenance would be easier since the data is already available. I could change it to use the query.

PR #565, opened 2025-05-26
[FSTORE-1752] mysql connection pool is not handled properly when the server closes connection

manu-sj:
The PR fixes an issue in which feature vector retrieval fails after a long period of inactivity.

**Root Cause:**
The [wait_timeout](https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_wait_timeout) variable inside MySQL specifies the time a connection can be idle before it is closed. By default that time is 8hrs (28800 secs). This causes the connections inside the connection pool to get disconnected.

This problem can be handled by using the [`pool_recycle`](https://aiomysql.readthedocs.io/en/stable/pool.html) argument which [recycles a connection before it becomes state](https://docs.sqlalchemy.org/en/20/core/pooling.html#setting-pool-recycle)

**Fix Done:**
Set the `pool_recycle` variable to half the default `wait_timeout` value.

JIRA Issue: https://hopsworks.atlassian.net/browse/FSTORE-1752

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/util_sql.py
+++ b/python/hsfs/core/util_sql.py
@@ -113,7 +113,7 @@ async def create_async_engine(
         loop=loop,
         minsize=options.get("minsize", default_min_size),
         maxsize=options.get("maxsize", default_min_size),
-        pool_recycle=options.get("pool_recycle", -1),
+        pool_recycle=options.get("pool_recycle", 14400),
         autocommit=options.get("autocommit", True),
     )
     return pool


SirOibaf:
@manu-sj - what happens if the mysql server is restarted before the 1440 is expired?

PR #598, opened 2025-07-15
Hopsworks mcp

ErmiasG:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


aversey:
During the merge of the repos, we had the idea that everything should be available from hopsworks directly. So I suggest moving it to `hopsworks.mcp`. `hsfs`, `hsml` are mainly kept for backwards-compatibility.

Also, why do we have tools wrapped in classes? What's the point of it? It is easier and simpler to just do:
```
@mcp.tool(tags=["auth"])
def login(...):
```
Plus, as I said, we should pay a lot of attention to typing, since it is propagated to the LLM as the schema of the tool.

For descriptions of tools, it is much better to follow Markdown plain text -- it is not the documentation of the Python functions, but of tools. The information about inputs and outputs is to be provided via the schemas. If you remember, I mentioned Pydantic models before.

Next, the Context param is _always_ provided by FastMCP. There is no need to check if it is null or not, and there is also no point in making it optional.

The prompts seem useless to me. The idea of MCP prompts is to help to build complex prompt templates, here the returned prompts are simpler than their descriptions. Resources, although useful, are to be called by the _user_ directly (same as prompts). They are _not_ called by the LLMs. For now, I would say to omit both prompts and resources and focus on tools.

For an easy way to start the server, we can add a command to the pyproject.toml, like this:
```
[project.scripts]
hopsworks-mcp = "hopsworks.mcp:main"
```
Then once hopsworks is installed, the users can simply run `hopsworks-mcp` as a command from their CLI, and it will run the main function from `hopsworks.mcp` package.

By the way, there should be an `__init__.py` in `hopsworks.mcp`.

Finally, the thing which worries me is that the current MCP interface is stateful. So each of these servers can be used by a single user only. I think the tools have to be reworked to be stateless, so that the tools assume they are already logged in to a project or login themselves each time a request is made. I think the first solution is the cleanest and to be preferred. In this case, the login should be done in the mcp server starting command.

python/hopsworks_mcp/tools/feature_store.py
```diff
@@ -0,0 +1,99 @@
...
+        Returns:
+            dict: A dictionary containing the feature store information or an error message.
+        """
+        if ctx:
+            await ctx.info(f"Retrieving feature store with name {name}...")
+        try:
+            project = hopsworks.get_current_project()
+            feature_store = project.get_feature_store(name=name)
+            if not feature_store:
+                return {
```
aversey, line 81: It is better to throw an Exception here.

python/hopsworks_mcp/version.py
aversey, line 1: It is not needed once we move the whole package to `hopsworks.mcp`.

aversey:
> I do not think putting all methods in one file is a good idea. When adding more tools it will get messy. And this will allow creating mcp server with only part of the tools or resources. So I will keep the wrapper classes.
> 
> As for the statefulness of the server I do not see any problem. We are not going to serve multiple users with one mcp server. A login method that uses an environment variable can be added and the auth tool can be disabled in that case.
> 
> The prompts and resources are just examples we will add more when we need them.

I have not meant to put all the tools in a single file. It is possible to register individual tools from multiple files, while they are defined as functions. I don't see a benefit of these classes and therefore think that this wrapping should be removed as redundant. It is also definitely possible to add or remove tools using tags (that's what they are for), so I don't think the point about creating partial MCP servers is valid.

I would be very careful with statefullness, as it may confuse the LLMs. As for the login tool, and prompts, and resources -- it is totally fine to keep them, but I don't see a clear benefit from them.

ErmiasG:
I think it is cleaner to keep them separate.

aversey (dismissed):
Ok, I think it can be merged. But I also think that it would be better to structure it differently (maybe by placing both the tools and the models in the main API structure, and just connect all of them in hopsworks.mcp (hopsworks.mcp can also be the place for utils); or by expanding each topic/tag into a package containing the corresponding tools in their own .py files together with the models -- in both cases the idea is to keep tightly connected things close) and remove prompts and resources just to keep it clean.

aversey (approved):
So, during the today's meeting, we agreed that the restructuring will be done as soon as a mistake / bug is made by forgetting to update the tool while modifying the method it wraps around; but for now we will follow the structure of this PR for the new tools.

PR #813, opened 2026-02-02
[FSTORE-1964] Upgrade Opensearch to 3.4

kennethmhc:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```markdown
```


ErmiasG:
FSTORE-1964 is now targeting OpenSearch 3.8.0 rather than 3.4, and everything of
substance in this PR has since landed on `main` through the FSTORE-1970 (2.19.5)
work:

- the `knn` query with an inner `filter` (replacing `knn` inside `bool.must`), in
  #1008 along with the k-too-large parsing
- the pyjks replacement (`_load_jks` / `_decrypt_jks_key`, `cryptography` in place
  of `pyjks`)
- `requires-python = ">=3.10,<3.14"`

The only remaining delta is `opensearch-py`, which is `>=1.1.0,<=2.8.0` on `main`
and `>=3.0.0,<=3.1.0` here, and that is not needed for 3.8.0 compatibility. Every
request form the 2.x client sends was checked against a live 3.8.0 cluster and
still answers 200: `knn` inside `bool.must`, a single non-array `must` object, and
the k-too-large and `Result window is too large` error bodies are unchanged in
shape. The loadtest `test_opensearch.py` workflow also passes against 3.8.0 on the
current pin.

So no `hopsworks-api` change is needed for the upgrade. The repos in the ticket are
docker-images #1020, hopsworks-helm #2317, hopsworks-ee #3332 and
clusterj-onlinefs #195. If the `opensearch-py` 3.x bump is still wanted it is worth
its own ticket, since it changes a user-facing dependency range and the base images
currently install `opensearch-py==2.8.0`.

PR #921, opened 2026-04-26
hsfs: accept Spark Connect DataFrames in ingest path

jimdowling:
## Summary

The Spark Connect client returns `pyspark.sql.connect.dataframe.DataFrame`, which is API-compatible with the classic `pyspark.sql.DataFrame` but a distinct class. Five `isinstance` checks in `hsfs/engine/spark.py` and one in `hsfs/core/schema_validation.py` rejected Connect DataFrames, breaking `fg.insert`, `fg.save`, `td.materialize` and every other write path that flows through `convert_to_default_dataframe`. The body of that method also called `dataframe.rdd`, which is unavailable in Connect mode.

## What changed

**Phase A — minimal fix to unblock ingest** (`3381dd9d5`):

- New `is_spark_dataframe()` in `hopsworks_common.spark_connect_utils` accepting both classic and Connect DataFrame classes (two separate `try`/`except ImportError` blocks so older PySparks without the `connect` submodule keep working).
- Replaced the six classic-only `isinstance` sites with the shared predicate.
- Replaced `createDataFrame(df.rdd, schema)` inside `convert_to_default_dataframe` with `df.to(schema)`. `DataFrame.to` is a PySpark 3.4+ API available identically on classic and Connect; same row-reconciliation semantics, no JVM/RDD round-trip.
- New unit tests for the predicate (classic, Connect via stub, pandas/list/None) and two regression tests on the convert path: schema is all-nullable after conversion, `DataFrame.rdd` is never accessed.
- New caveat doc at `.claude/docs/caveats/spark-connect.md` summarizing what works/doesn't in Connect mode.

**Phase B — audit + docstrings** (`ce7c21604`):

- Audited every JVM/RDD access in hsfs: `engine/spark.py`, `core/hudi_engine.py`, `core/delta_engine.py`, `core/kafka_engine.py`, `storage_connector.py`. All already gated on `self._is_connect` with fallbacks or fail-fast errors. **No functional changes needed.**
- Added authoritative docstrings to `Engine.convert_to_default_dataframe` and `Engine.is_spark_dataframe` documenting the Connect path.
- Left the 66 `TypeVar("pyspark.sql.DataFrame")` markers in public APIs unchanged — they are consumed by `hopsworks_apigen` for doc generation, and from the user's perspective "Spark DataFrame" already covers both modes through the shared predicate.

## Test plan

- [x] `pytest python/tests/test_spark_connect_utils.py` — all predicate tests pass (classic, Connect via stub, pandas/list/None).
- [x] `pytest python/tests/engine/test_spark.py -k convert_to_default_dataframe` — existing + new regression tests pass.
- [x] `ruff check` — clean.
- [x] `docsig hopsworks_common hsfs` — clean.
- [ ] Manual: from inside `terminal-spark` (Spark Connect mode), build a DataFrame with `spark.range(10).withColumn(...)` and call `fg.save(df)` / `fg.insert(df)`. Confirm rows land in the offline store.

## Why two commits, one PR

Phase A is the minimum surgical fix; Phase B is the audit + docstring follow-up. Both are tightly scoped to the same area and ship together to give reviewers the full context.

### DIFF
--- a/.claude/docs/caveats/spark-connect.md
+++ b/.claude/docs/caveats/spark-connect.md
@@ -1,0 +1,29 @@
+# Spark Connect compatibility
+
+`hsfs/engine/spark.py` runs in two modes:
+
+- **Classic** — JVM-resident Spark, full `SparkContext`, `_jvm`, `.rdd`, etc.
+- **Spark Connect** — gRPC client-only session. `SparkSession.builder.getOrCreate()` returns a thin client, DataFrames are `pyspark.sql.connect.dataframe.DataFrame`, no JVM access on the driver.
+
+The engine detects Connect mode at startup via `is_spark_connect_session()` and exposes `self._is_connect`. JVM-touching code paths are gated on that flag; the predicate `is_spark_dataframe()` (in `hopsworks_common/spark_connect_utils.py`) is the single source of truth for "is this a Spark DataFrame I can operate on?" — it accepts both classic and Connect classes.
+
+## Fix for the convert path
+
+The original `convert_to_default_dataframe` used `self._spark_session.createDataFrame(df.rdd, nullable_schema)` to make every field nullable. `.rdd` is unavailable in Connect mode. The replacement is `df.to(nullable_schema)` (PySpark 3.4+) — same row-reconciliation semantics, no JVM round-trip, works in both modes.
+
+## What's known to work in Connect
+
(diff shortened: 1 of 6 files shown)


jimdowling:
Folded into #920 — both spark-connect commits (`hsfs: accept Spark Connect DataFrames in ingest path` + `hsfs: document Spark Connect support in engine docstrings`) cherry-picked onto the `cli` branch in commits 1af9a35e9 and 4e3d09ad9. Source branch will be deleted.

PR #949, opened 2026-05-12
[HWORKS-2780] Add Python Agent Image

jimdowling:
## Summary

- Pulls in @aversey's `agent-deployments` work (`ms.deploy_agent(...)`), which lets a user deploy a local `.py` script or `pyproject.toml`-rooted package as a server-only Hopsworks deployment backed by the new `python-agent-pipeline` base image.
- Adds a `hops agent` CLI command group mirroring `hops deployment`: `list | info | create | start | stop | query | logs | delete`. Registered lazily in `cli/main.py` so cold paths (`hops --help`, `hops setup`) don't pay the SDK import cost.

## Test plan

- [x] `ruff check --config python/pyproject.toml` on every changed `.py` — passes
- [x] `ruff format --check` on every changed `.py` — passes (121 files already formatted)
- [x] `docsig hopsworks_common hopsworks hsfs hsml` — `agent.py` clean; only pre-existing SIG306 *warnings* elsewhere
- [x] `hops agent --help`, `hops agent create --help`, `hops agent query --help`, `hops agent logs --help` render correctly (verified with Click's `CliRunner`)
- [ ] CI matrix `pytest python/tests` (left to CI — local `uv sync` is blocked by a separate `hopsworks_apigen` griffe alias-resolution issue when building the editable wheel; unrelated to this CLI change)

### DIFF
--- a/python/hopsworks/cli/commands/agent.py
+++ b/python/hopsworks/cli/commands/agent.py
@@ -1,0 +1,408 @@
+"""``hops agent`` — Python agent deployment lifecycle.
+
+Agents are server-only deployments (no model attached) created from a local
+``.py`` script or a directory with ``pyproject.toml``.
+``create`` calls ``ms.deploy_agent`` which builds/uploads the code, provisions
+a Python environment from the ``python-agent-pipeline`` base image, and
+creates or updates a regular Hopsworks deployment.
+The remaining verbs (``info``, ``start``, ``stop``, ``query``, ``logs``,
+``delete``) mirror ``hops deployment`` but resolve names through an
+agent-only lookup that excludes model-backed deployments.
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import Any
+
+import click
+from hopsworks.cli import output, session
+
+
+@click.group("agent")
+def agent_group() -> None:
+    """Python agent deployment commands."""
+
+
+@agent_group.command("list")
+@click.pass_context
+def agent_list(ctx: click.Context) -> None:
+    """List every agent deployment in the active project.
+
+    Args:
+        ctx: Click context.
+    """
+    ms = _get_model_serving(ctx)
+    try:
+        deployments = ms.get_deployments()
+    except Exception as exc:  # noqa: BLE001
+        raise click.ClickException(f"Could not list agents: {exc}") from exc
+
+    rows = []
+    for d in deployments or []:
+        if not _is_agent(d):
+            continue
+        rows.append(
+            [
+                getattr(d, "id", "?"),
+                getattr(d, "name", "?"),
+                _predictor_attr(d, "environment", "-"),
+                _predictor_attr(d, "script_file", "-"),
+                _deployment_status(d),
+            ]
+        )
+    output.print_table(["ID", "NAME", "ENVIRONMENT", "SCRIPT", "STATUS"], rows)
+
+
+@agent_group.command("info")
+@click.argument("name")
+@click.pass_context
+def agent_info(ctx: click.Context, name: str) -> None:
+    """Show details for a single agent.
+
+    Args:
+        ctx: Click context.
+        name: Agent name.
+    """
+    agent = _get_agent(ctx, name)
+
+    if output.JSON_MODE:
+        output.print_json(_agent_to_dict(agent))
+        return
+
+    rows = [
+        ["ID", getattr(agent, "id", "?")],
+        ["Name", getattr(agent, "name", "?")],
+        ["Environment", _predictor_attr(agent, "environment", "-")],
+        ["Script", _predictor_attr(agent, "script_file", "-")],
+        ["Serving tool", getattr(agent, "serving_tool", "-")],
+        ["Model server", getattr(agent, "model_server", "-")],
+        ["API protocol", getattr(agent, "api_protocol", "-")],
+        ["Description", getattr(agent, "description", "-") or "-"],
+        ["Status", _deployment_status(agent)],
+    ]
+    output.print_table(["FIELD", "VALUE"], rows)
+
+
+@agent_group.command("create")
+@click.argument("entry", type=click.Path(exists=True))
+@click.option("--name", help="Agent name; defaults to the basename of ENTRY.")
+@click.option(
+    "--requirements",
+    type=click.Path(exists=True),
+    help="Local requirements.txt to install into the agent environment.",
+)
+@click.option(
+    "--environment",
+    help="Python environment name; defaults to the agent name.",
+)
+@click.option(
+    "--upload-dir",
+    default="Resources/agents",
+    show_default=True,
+    help="HopsFS directory under which agent files are placed.",
+)
+@click.option("--description", default=None, help="Agent description.")
+@click.pass_context
+def agent_create(
+    ctx: click.Context,
+    entry: str,
+    name: str | None,
+    requirements: str | None,
+    environment: str | None,
+    upload_dir: str,
+    description: str | None,
+) -> None:
+    """Create (or update) an agent from a local script or package.
+
+    ``ENTRY`` is a ``.py`` file or a directory containing ``pyproject.toml``.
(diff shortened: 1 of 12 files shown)


python/hopsworks/cli/commands/agent.py
```diff
@@ -0,0 +1,408 @@
...
+    """
+    ms = _get_model_serving(ctx)
+    try:
+        deployments = ms.get_deployments()
+    except Exception as exc:  # noqa: BLE001
+        raise click.ClickException(f"Could not list agents: {exc}") from exc
+
+    rows = []
+    for d in deployments or []:
+        if not _is_agent(d):
```
javierdlrm, line 44: This should be a query parameter in the URI that retrieves only agents from the DB

python/hopsworks/cli/commands/agent.py
```diff
@@ -0,0 +1,408 @@
...
+    """
+    agent = _get_agent(ctx, name)
+    try:
+        agent.stop(await_stopped=wait)
+    except Exception as exc:  # noqa: BLE001
+        raise click.ClickException(f"Stop failed: {exc}") from exc
+    output.success("✓ Stopped agent %s", name)
+
+
+@agent_group.command("query")
```
javierdlrm, line 201: Is `query` the appropriate verb? Compared to: 
- `prompt` -- 
- `call` -- like in MCP specs
- `request` -- direct command for an agent to do something

I think that we can have two commands, similar to model deployments. One for flexible payloads, another one for prompts. The easy-by-default allow-for-complex principle. For example,
- `request` or `call` -- accepts any JSON as `--data `
- `prompt` -- accepts a string, we built the {"prompt": "hello"} as in the example.

It's more natural to call `agent prompt "Hello!"` than `agent query {"prompt": "Hello!"}`

python/hopsworks/cli/commands/agent.py
```diff
@@ -0,0 +1,408 @@
...
+        raise click.ClickException(f"Delete failed: {exc}") from exc
+    output.success("✓ Deleted agent %s", name)
+
+
+def _get_model_serving(ctx: click.Context) -> Any:
+    project = session.get_project(ctx)
+    return project.get_model_serving()
+
+
+def _get_agent(ctx: click.Context, name: str) -> Any:
```
javierdlrm, line 368: Not necessary if we add the query param to filter in the backend

python/hopsworks/cli/commands/agent.py
```diff
@@ -0,0 +1,408 @@
...
+    except Exception as exc:  # noqa: BLE001
+        raise click.ClickException(f"Agent '{name}' not found: {exc}") from exc
+    if d is None or not _is_agent(d):
+        raise click.ClickException(f"Agent '{name}' not found.")
+    return d
+
+
+def _is_agent(d: Any) -> bool:
+    """Identify server-only (agent) deployments: no model attached."""
+    return getattr(d, "model_name", None) in (None, "")
```
javierdlrm, line 389: Is that the only check for classifying deployments as agents? What about Python deployments and Apps?
We should probably add a type column in a follow-up PR.

python/hsml/model_serving.py
```diff
@@ -16,18 +16,33 @@
...
 
 import os
+import posixpath
+import re
+import tempfile
+import zipfile
 from typing import TYPE_CHECKING
 
+from build import ProjectBuilder
+from build.env import DefaultIsolatedEnv
```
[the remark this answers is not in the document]
javierdlrm: I think it's a good idea to move these imports inside the `_build_and_install_package()` method

python/hsml/model_serving.py
```diff
@@ -397,6 +412,139 @@ def create_endpoint(
...
+        if is_script:
+            script_file = ds_api.upload(entry_abs, agent_dir, overwrite=True)
+        else:
+            script_file = _build_and_install_package(ds_api, env, entry_abs, agent_dir)
+        # The serving backend expects the script path under /Projects/<proj>/...
+        script_file = util.convert_to_abs(script_file, self._project_name)
+
+        if requirements is not None:
+            req_remote = ds_api.upload(
+                os.path.abspath(requirements), agent_dir, overwrite=True
```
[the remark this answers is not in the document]
javierdlrm: That's actually true, no checks for existing requirements file

python/hsml/model_serving.py
```diff
@@ -494,3 +642,133 @@ def project_id(self):
 
     def __repr__(self):
         return f"ModelServing(project: {self._project_name!r})"
+
+
+_AGENT_IDENTIFIER_RE = re.compile(r"^[A-Za-z0-9_-]+$")
```
javierdlrm: We normally place constants at the top of the file for visibility

python/hopsworks/cli/commands/agent.py
```diff
@@ -0,0 +1,408 @@
...
+
+@click.group("agent")
+def agent_group() -> None:
+    """Python agent deployment commands."""
+
+
+@agent_group.command("list")
+@click.pass_context
+def agent_list(ctx: click.Context) -> None:
+    """List every agent deployment in the active project.
```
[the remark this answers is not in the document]
javierdlrm: This is legit. Can you please add some tests for the `hops agent`?

javierdlrm (changes requested):

python/hsml/model_serving.py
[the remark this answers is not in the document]
jimdowling: fixing

python/hopsworks/cli/commands/agent.py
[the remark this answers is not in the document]
jimdowling: ack

python/hopsworks/cli/commands/agent.py
jimdowling: We can do this in a later PR - requires a new method in hopsworks-ee

python/hopsworks/cli/commands/agent.py
jimdowling: I use the term 'query' in the book. You prompt a LLM but query an agent. 

I am forward thinking - the query will also include API params, won't just be a text prompt.

jimdowling: leaving as query

python/hopsworks/cli/commands/agent.py
jimdowling: A follow-up PR would be my preference. need to get this in. 
Here is what claude says about the name:

  ---
  Reply to "Is model_name the only check? What about Python deployments and Apps?"
  
  Apps aren't an issue — project.create_app(...) goes through /project/{id}/jobs/{name} with type=pythonAppJobConfiguration, so Streamlit/Python apps never show up in ms.get_deployments().

  Python deployments are different though, and you're right to flag it. The SDK exposes Predictor.for_server(name, script_file, …) (and ms.create_predictor(script_file=…, model=None)) which
  produce server-only deployments with model_name=None / has_model=False — exactly what _is_agent matches today. So any non-agent server-only deployment created via that public API would be
  listed as an "agent" by the CLI.

  In practice the only first-party caller of that path is deploy_agent, but the API surface allows it, so this is structurally fragile.

  Agreed on the follow-up: add an explicit kind/type column on serving (e.g. MODEL / AGENT / SCRIPT), set it server-side based on the request shape, and have the SDK filter on that. I'll
  file a follow-up JIRA — happy to keep this PR's behavior (filter on has_model) for now and tighten in the follow-up.

  As a stopgap if you'd rather not ship the loose heuristic at all, we can also gate on upload_dir (Resources/agents/<name>/) which deploy_agent always uses, but that breaks if a user
  customizes --upload-dir. I'd lean toward leaving the has_model check as-is and tracking the proper fix.

python/hopsworks/cli/commands/agent.py
javierdlrm: I think they are not exclusive. We can use `query`, but also add a `prompt` command as a shortcut for prompts without API params.

javierdlrm (approved):

PR #959, opened 2026-05-21
[FSTORE-2030] Add support for specifying lookback windows for PIT queries

manu-sj:
## Summary

- Adds a `Lookback` dataclass under `hsfs.constructor.lookback` and a `lookback` parameter on `FeatureView.get_batch_data`, `create_training_data`, and the split variants. The window applies uniformly to every joined feature group in the PIT join.
- Two modes: `partition_key` (constant-bound predicate on each FG's partition column — flyingduck and Spark prune partitions before reading files), `event_time` (predicate on the FG's `event_time` column — row-level correctness; partition pruning is engine-dependent).
- Wire format is camelCase (`lookbackKey`, `startWindow`, `endWindow`) with bounds as epoch ms; the SDK fans the uniform lookback out across the query tree's joins and also emits a `joinsLookback` map keyed by `prefix + fg.name` on the training-dataset POST.

## JIRA

[FSTORE-2030](https://hopsworks.atlassian.net/browse/FSTORE-2030)

## Test plan

- [x] 27 unit tests on `Lookback` (validation matrix, dict/dataclass round-trip, date+datetime coercion, nested-map serialization).
- [x] `feature_view` and `training_dataset` tests pass.
- [x] `ruff` + `docsig` clean on the new module and the modified public methods.
- [x] End-to-end on cluster `manu-test` via the `loadtest` integration tests (companion PR).

## Companion PRs

- Backend: `logicalclocks/hopsworks-ee` → branch `FSTORE-2030`
- Integration tests: `logicalclocks/loadtest` → branch `FSTORE-2030`
- Docs: `logicalclocks/logicalclocks.github.io` → branch `FSTORE-2030`

[FSTORE-2030]: https://hopsworks.atlassian.net/browse/FSTORE-2030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/hsfs/constructor/lookback.py
+++ b/python/hsfs/constructor/lookback.py
@@ -1,0 +1,431 @@
+#
+#   Copyright 2026 Logical Clocks AB
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+from hopsworks_apigen import public
+from hopsworks_common.util import convert_event_time_to_timestamp
+from hsfs.decorators import typechecked
+
+
+if TYPE_CHECKING:
+    from datetime import date, datetime
+
+
+EVENT_TIME = "event_time"
+PARTITION_KEY = "partition_key"
+_VALID_LOOKBACK_KEYS = (EVENT_TIME, PARTITION_KEY)
+
+_UNIFORM_LOOKBACK_DICT_KEYS = frozenset({"key", "start", "end"})
+_LOOKBACKS_DICT_KEYS = frozenset({"default", "feature_groups"})
+
+
+@public
+@typechecked
+class Lookback:
+    """A lookback window for PIT joins on a feature view.
+
+    Bounds the rows of the joined feature group(s) that participate in the
+    point-in-time join, so flyingduck and Spark can prune partitions before
+    reading files.
+
+    `key="partition_key"` emits the predicate against the FG's partition
+    column — flyingduck and Spark prune partitions before reading files.
+    `key="event_time"` emits against the event_time column — row-level
+    correctness only; partition pruning is engine-dependent.
+
+    Use this class for a uniform window across every FG; use
+    [`Lookbacks`][hsfs.constructor.lookback.Lookbacks] for per-FG
+    overrides.
+
+    Example:
+        ```python
+        fv.get_batch_data(
+            lookback=Lookback(
+                key="partition_key",
+                start=date(2026, 5, 5),
+                end=date(2026, 5, 17),
+            ),
+        )
+        ```
+    """
+
+    def __init__(
+        self,
+        key: str,
+        start: date | datetime | int | str,
+        end: date | datetime | int | str | None = None,
+    ) -> None:
+        if key not in _VALID_LOOKBACK_KEYS:
+            raise ValueError(
+                f"Lookback `key` must be one of {_VALID_LOOKBACK_KEYS!r}; got {key!r}."
+            )
+        if start is None:
+            raise ValueError("Lookback `start` is required.")
+        if end is not None:
+            start_ms = convert_event_time_to_timestamp(start)
+            end_ms = convert_event_time_to_timestamp(end)
+            if start_ms >= end_ms:
+                raise ValueError(
+                    f"Lookback `start` ({start!r}) must be strictly earlier "
+                    f"than `end` ({end!r})."
+                )
+        self._key = key
+        self._start = start
+        self._end = end
+
+    @property
+    def key(self) -> str:
+        """Column the lookback predicate targets on each joined feature group."""
+        return self._key
+
+    @property
+    def start(self) -> date | datetime | int | str:
+        """Required lower bound of the lookback window.
+
+        Holds whatever shape `__init__` accepted: a `date`/`datetime` for user-form
+        construction, or epoch-millis `int` / ISO `str` after `from_response_json`
+        reconstructs from the wire payload. `convert_event_time_to_timestamp` accepts
+        all four shapes uniformly.
+        """
+        return self._start
+
+    @property
+    def end(self) -> date | datetime | int | str | None:
+        """Optional upper bound; `None` emits a one-sided lower-bound predicate only.
+
+        Same value-shape contract as `start`.
+        """
+        return self._end
+
+    def to_dict(self) -> dict[str, Any]:
+        """Return the wire-format payload for the backend.
+
+        Bounds are emitted as epoch milliseconds; `key` is upper-cased to
+        match the backend enum; `end` is omitted when not provided.
+
+        Returns:
+            The wire-format dict with keys `lookbackKey`, `start`, and
+            optionally `end`.
+        """
+        payload: dict[str, Any] = {
+            "lookbackKey": self._key.upper(),
+            "start": convert_event_time_to_timestamp(self._start),
+        }
+        if self._end is not None:
+            payload["end"] = convert_event_time_to_timestamp(self._end)
+        return payload
+
+    @classmethod
+    def from_dict(cls, value: dict[str, Any] | Lookback | None) -> Lookback | None:
+        """Construct a `Lookback` from a user-form dict (or pass through an instance).
+
+        Accepts `{"key", "start", "end"}`, an existing `Lookback`, or `None`.
+
+        Args:
+            value: A user-form dict, an existing `Lookback`, or `None`.
+
+        Returns:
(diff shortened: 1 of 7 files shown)


python/hsfs/constructor/lookback.py
```diff
@@ -0,0 +1,431 @@
+#
+#   Copyright 2026 Logical Clocks AB
```
javierdlrm:
```suggestion
#   Copyright 2026 Hopsworks AB
```

python/hsfs/constructor/lookback.py
```diff
@@ -0,0 +1,431 @@
...
+from hopsworks_common.util import convert_event_time_to_timestamp
+from hsfs.decorators import typechecked
+
+
+if TYPE_CHECKING:
+    from datetime import date, datetime
+
+
+EVENT_TIME = "event_time"
+PARTITION_KEY = "partition_key"
```
javierdlrm: Can we use upper case?

python/hsfs/constructor/lookback.py
```diff
@@ -0,0 +1,431 @@
...
+_LOOKBACKS_DICT_KEYS = frozenset({"default", "feature_groups"})
+
+
+@public
+@typechecked
+class Lookback:
+    """A lookback window for PIT joins on a feature view.
+
+    Bounds the rows of the joined feature group(s) that participate in the
+    point-in-time join, so flyingduck and Spark can prune partitions before
```
javierdlrm: flyingduck is an internal name. Let's use Hopsworks Query Service?

python/hsfs/constructor/lookback.py
```diff
@@ -0,0 +1,431 @@
...
+    reading files.
+
+    `key="partition_key"` emits the predicate against the FG's partition
+    column — flyingduck and Spark prune partitions before reading files.
```
javierdlrm: flyingduck is an internal name. Let's use Hopsworks Query Service?

python/hsfs/constructor/lookback.py
```diff
@@ -0,0 +1,431 @@
...
+        return self._key
+
+    @property
+    def start(self) -> date | datetime | int | str:
+        """Required lower bound of the lookback window.
+
+        Holds whatever shape `__init__` accepted: a `date`/`datetime` for user-form
+        construction, or epoch-millis `int` / ISO `str` after `from_response_json`
+        reconstructs from the wire payload. `convert_event_time_to_timestamp` accepts
+        all four shapes uniformly.
```
javierdlrm: I think we can make the docstring more user-friendly. No need to describe internal code (__init__, from_response_json, ...), but high-level purpose of "start" property.

python/hsfs/constructor/lookback.py
```diff
@@ -0,0 +1,431 @@
...
+        self._key = key
+        self._start = start
+        self._end = end
+
+    @property
+    def key(self) -> str:
+        """Column the lookback predicate targets on each joined feature group."""
+        return self._key
+
+    @property
```
javierdlrm, line 105: Do we need the `@public ` annotation on these properties? or are they not needed because the class is already "@public"?

python/hsfs/constructor/lookback.py
```diff
@@ -0,0 +1,431 @@
...
+        return cls(
+            key=str(key).lower(),
+            start=value.get("start"),
+            end=value.get("end"),
+        )
+
+
+@public
+@typechecked
+class Lookbacks:
```
javierdlrm: Not a hard requirement, but in my humble opinion these names are probably more descriptive for users / ai tools:

- `FeatureGroupLookback` instead of `Lookback`, to refer to a lookback applied in a specific feature group.
- `Lookback` instead of `Lookbacks`, to refer to the lookback of a FV query. This lookback contains multiple `FeatureGroupLookback` items

Then the code would look like:
```python
fv.get_batch_data(
    lookback=Lookback(
        default=FeatureGroupLookback(key="partition_key", start=date(2026, 5, 5)),
        feature_group_lookbacks={
            "transactions": FeatureGroupLookback(key="event_time",
                                       start=datetime(2026, 5, 1, tzinfo=timezone.utc)),
        },
    ),
)
```

A bit more verbose but I think it's clearer. If "feature_group_lookbacks" parameter looks to long, we could name it "overrides", but I think it's ok.

python/hsfs/constructor/query.py
```diff
@@ -761,6 +770,18 @@ def from_response_json(cls, json_dict: dict[str, Any]) -> Query:
...
         )
+        # Restore Lookbacks from the wire payload. Without this, any consumer
+        # that reconstructs a Query from JSON (notably the materialization
+        # Spark job that runs `Query.from_response_json` on the td_app_conf
+        # the SDK posted) would drop the lookback and run an unbounded PIT
+        # join. The SDK's local-process callers (get_batch_data, in-memory
+        # training_data) attach `_lookbacks` themselves after this method
+        # returns, so the only behavior change here is for cross-process
+        # reconstructions.
+        from hsfs.constructor.lookback import Lookbacks
```
javierdlrm: Do we need the import here? It is already in [line 41](https://github.com/logicalclocks/hopsworks-api/pull/959/changes#diff-43d6c9878bedb3426c17d586da74bcf174cb4b56b405768eca3d442b1b3f7038R41)

python/hsfs/core/feature_view_engine.py
```diff
@@ -52,6 +52,7 @@
 if TYPE_CHECKING:
     import pandas as pd
     import polars as pl
+    from hsfs.constructor import lookback as _lookback
```
javierdlrm: There are two different items with same "_lookback" name used in this file:
- `import lookback as _lookback`
- `query._lookback`

It seems that the "_lookback" import is only used for accessing the `_lookback.Lookbacks` class. Could we just import this class directly here?  or is there a reason why it was not imported directly (e.g., cyclic dependency)?

python/hsfs/core/feature_view_engine.py
```diff
@@ -446,11 +457,31 @@ def create_training_dataset(
...
+        if lookback is not None:
+            training_dataset_obj._lookbacks = lookback
         self._set_event_time(feature_view_obj, training_dataset_obj)
         updated_instance = self._create_training_data_metadata(
             feature_view_obj, training_dataset_obj
         )
+        # `_create_training_data_metadata` calls
+        # `TrainingDataset.update_from_response_json` which re-runs
+        # `__init__(**response)` and silently resets `_lookbacks` when the
+        # backend response omits the field. Re-set it on both the local TD
```
javierdlrm: Shouldn't the backend response include the lookback field if set in the previous _create_training_data_metadata call?

javierdlrm (changes requested):

python/hsfs/constructor/lookback.py
```diff
@@ -0,0 +1,431 @@
...
+        ```
+    """
+
+    def __init__(
+        self,
+        key: str,
+        start: date | datetime | int | str,
+        end: date | datetime | int | str | None = None,
+    ) -> None:
+        if key not in _VALID_LOOKBACK_KEYS:
```
javierdlrm: We could add a user-friendly "key.upper()" here.

python/hsfs/constructor/lookback.py
manu-sj: Updated License in all files to use Hopsworks

python/hsfs/constructor/lookback.py
manu-sj: Set to uppercase and updated the docs too use Uppercase too

python/hsfs/constructor/lookback.py
manu-sj: Replaced all occurrence of flyingduck with Hopsworks Query Service. In this case I updated the comment to remove mention of spark and arrowflight since they were internal implementation details

python/hsfs/constructor/lookback.py
manu-sj: Replaced all occurrence of flyingduck with Hopsworks Query Service. In this case I updated the comment to remove mention of spark and arrowflight since they were internal implementation details

python/hsfs/constructor/lookback.py
manu-sj: Updated docs to remove internal implemenation details

python/hsfs/constructor/lookback.py
manu-sj: Yes you are right I missed this updated properties to include `@public` decorator

python/hsfs/constructor/lookback.py
manu-sj: I agree updated the Naming and also backend and docs

python/hsfs/constructor/query.py
manu-sj: Import at line 41, was under `TYPE_CHECKING`, but in this case we do not have circular dependencies so I moved it to the top outside the TYPE_CHECKING section.

python/hsfs/core/feature_view_engine.py
manu-sj: Fix it .

Moved import to be `from hsfs.constructor.lookback import Lookback`

`query._lookback`  was a internal variable, so I created property for it so that private variable are not accessed outside the class.

python/hsfs/constructor/lookback.py
manu-sj: Casting done

javierdlrm (approved):

PR #966, opened 2026-05-21
[FSTORE-2036] PR 4a/4 — Python SDK support for Unity Catalog OAuth M2M

jimdowling:
## Summary
PR 4 of 4 for FSTORE-2099, **hopsworks-api half**. Extends `UnityCatalogConnector` so the Python SDK round-trips the new OAuth fields the backend (#3032 / #3033) and frontend (logicalclocks/hopsworks-front#1919) added. Legacy PAT-only construction keeps working unchanged.

Companion loadtest PR: logicalclocks/loadtest (link in this thread once the loadtest PR is opened).

Spec: `uc-oauth2/uc-oauth2.md` in the per-feature workspace. Ticket: https://hopsworks.atlassian.net/browse/FSTORE-2099

### What changes
- **Constructor** gains `auth_method`, `client_id`, `client_secret`, `oauth_endpoint`, `account_id`, `account_host`, `has_access_token`, `has_client_secret`. `auth_method` defaults to `"PAT"` when absent so existing code paths (and downstream fixtures) keep producing PAT connectors. `OAUTH_M2M` without an explicit `oauth_endpoint` defaults to `"WORKSPACE"`, matching the frontend default.
- **Write-only-friendly booleans**: `has_access_token` and `has_client_secret` come from the server (`hasAccessToken` / `hasClientSecret` in camelCase). When a caller builds a connector locally with a secret in hand, the booleans fall back to "is the secret non-None" so client code that constructs in-process still reports correct state.
- **`from_response_json` is unchanged** — it already uses `humps.decamelize` + `**kwargs` splat, which picks up the new fields by name once they're declared on the constructor.
- **Existing `get_unity_catalog` fixture updated** to match the post-PR-1 backend wire format (`hasAccessToken: true`; no decrypted `access_token` in GET responses). Two new fixtures (`get_unity_catalog_oauth_workspace`, `get_unity_catalog_oauth_account`) cover the OAuth modes.
- **Tests** extended from 4 to 8 in `TestUnityCatalogConnector` — round-trip for both OAuth modes; legacy construction defaulting to PAT; OAuth construction defaulting `oauth_endpoint` to `WORKSPACE`.

### Test plan
- [x] `uv run pytest python/tests/test_storage_connector.py::TestUnityCatalogConnector` — **8/8 passing**.
- [x] `uv run ruff check` — clean.
- [x] `uv run docsig python/hsfs/storage_connector.py` — clean.

### DIFF
--- a/python/hsfs/storage_connector.py
+++ b/python/hsfs/storage_connector.py
@@ -3679,10 +3679,7 @@ def _resolve_uc_spark_options(
         "Content-Type": "application/json",
     }
 
-    table_url = (
-        f"{workspace}/api/2.1/unity-catalog/tables/"
-        f"{catalog}.{schema}.{table}"
-    )
+    table_url = f"{workspace}/api/2.1/unity-catalog/tables/{catalog}.{schema}.{table}"
     table_resp = requests.get(table_url, headers=headers, timeout=30)
     _raise_for_uc_status(table_resp, "table lookup")
     table_payload = table_resp.json()
@@ -3698,9 +3695,7 @@ def _resolve_uc_spark_options(
             f"{data_source_format or 'unknown'}."
         )
 
-    creds_url = (
-        f"{workspace}/api/2.0/unity-catalog/temporary-table-credentials"
-    )
+    creds_url = f"{workspace}/api/2.0/unity-catalog/temporary-table-credentials"
     creds_resp = requests.post(
         creds_url,
         headers=headers,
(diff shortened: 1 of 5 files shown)


python/hsfs/storage_connector.py
```diff
@@ -3365,13 +3365,49 @@ def __init__(
         default_catalog: str | None = None,
         aws_region: str | None = None,
         arguments: list[dict[str, Any]] | dict[str, Any] | None = None,
+        auth_method: str | None = None,
```
bubriks, line 3378: can this connector be used both in python and pyspark? Right now these changes just look like updated metadata but it doesnt look to be used in the api.

jimdowling: I will add support for pyspark

jimdowling: Adding pyspark, forgot about it.

bubriks (approved):

PR #997, opened 2026-06-09
[HWORKS-2849] Make hopsworks-api follow PEP 8: @public surface + privatize internals

robzor92:
## What

Makes `hopsworks-api` follow PEP 8: a symbol is part of the public API **iff** it is annotated `@public` (from `hopsworks_apigen`); everything else is an internal implementation detail and is now `_`-prefixed. This lets tooling (and humans) rely on the underscore convention instead of guessing.

Driver: an internal agent reads the source and calls non-`@public` functions because they don't look private. After this change, internal API is unambiguously `_`-prefixed.

## Phases

**Phase 1 — complete the `@public` surface** (`22d4285f5`)
`@public` was incomplete, so a mechanical "underscore everything" would have amputated real public API. Added authoritative `@public` coverage across `hsfs`/`hsml`/`hopsworks_common` (domain entities, `project.get_*_api()` handles, catchable exceptions, builtin transformations, etc.), validated against the loadtest consumer + the docs cross-references.

**Phase 2 — privatize internals** (4 `[APPEND]` commits)
Underscore-prefixed internal methods across all engine modules and updated every call site + test:
- `01bbadf4c` — 18 `hsfs/core/*_engine.py` + `hsml/engine/{local,model,serving}`
- `7fe0da3f3` — `hsfs/engine` compute engines (`python`/`spark` **polymorphic interface** renamed consistently across impls + all 108 `engine.get_instance().X` dispatch sites)
- `7f5b19f1a` — `transformation_function_engine`
- `319de9716` — `online_store` sql/rest engines

**Carve-outs (left public):** serialization called by-name by the framework (`json`/`to_dict`/`from_response_json`/…), `@property` accessors backing onto `self._field`, and methods overriding a `@public` abstract (polymorphism).

## Testing

Full suite green locally **excluding** `tests/engine/test_spark.py` (needs a JVM/Avro env that isn't reliable on my box): `2724 passed`. **Relying on CI to run `test_spark.py`** — the compute-engine batch touches `spark.py` heavily, so please watch that job.

## Follow-up (separate PR)

A CI lint gate in `hopsworks-apigen` (on its existing `HopsworksApigenGriffe` extension, which already tags `is_public`) to enforce "non-`@public` ⇒ `_`-prefixed" going forward.


python/hopsworks_common/client/exceptions.py
```diff
@@ -215,31 +207,25 @@ class DatasetException(Exception):
...
 class OpenSearchException(Exception):
     """Generic opensearch exception."""
 
 
-@also_available_as(
-    "hopsworks.client.exceptions.JobExecutionException",
-    "hsml.client.exceptions.JobExecutionException",
-)
+@public("hopsworks.client.exceptions.JobExecutionException")
+@also_available_as("hsml.client.exceptions.JobExecutionException")
```
manu-sj, line 223: Not really something big but can these be

`hopsworks.hsml.client.exceptions.JobExecutionException` instead of `hsml.client.exceptions.JobExecutionException`. 

We were at some point thinking of removing direct reference to `hsml` and instead use it though `hopsworks` 

Same for the below ones too

python/hopsworks_common/core/dataset_api.py
```diff
@@ -1066,6 +1069,7 @@ def zip(
 
     # region Dataset Tags
 
+    @public
```
manu-sj, line 1072: Shouldn't this be a @deprecated - i think we have a tags API for this

python/hopsworks_common/core/dataset_api.py
```diff
@@ -1093,6 +1097,7 @@ def add(self, path: str, name: str, value: str):
         json_value = json.dumps(value)
         _client._send_request("PUT", path_params, headers=headers, data=json_value)
 
+    @public
```
manu-sj, line 1100: Shouldn't this be a @deprecated - i think we have a tags API for this

python/hopsworks_common/core/dataset_api.py
```diff
@@ -1116,6 +1121,7 @@ def delete(self, path: str, name: str):
         ]
         _client._send_request("DELETE", path_params)
 
+    @public
```
manu-sj, line 1124: Shouldn't this be a @deprecated - i think we have a tags API for this

python/hopsworks_common/core/flink_cluster_api.py
```diff
@@ -30,7 +30,8 @@ class FlinkClusterApi:
     def __init__(self):
```
manu-sj: I think the flink API itself has to be removed. I don't think we support Flink in Hopsworks anymore.

python/hopsworks_common/core/rest.py
```diff
@@ -16,20 +16,20 @@
 """Stable internal REST helper for SDK consumers (CLI and ``core/<entity>_api`` modules).
```
manu-sj, line 16: I am not really sure why we need this module but we ignore it for now 😅

python/hopsworks_common/flink_cluster.py
```diff
@@ -71,7 +71,7 @@ def _count_ongoing_executions(self):
         return ongoing_executions
```
manu-sj: I think this is also a file that has to be removed Hopsworks does not have Flink inside anymore.

python/hsfs/embedding.py
```diff
@@ -164,14 +164,6 @@ def name(self):
         """str: The name of the embedding feature."""
         return self._name
 
-    @property
```
manu-sj, line 167: I think this should be kept and marked @public

python/hsfs/feature_group_writer.py
```diff
@@ -28,13 +30,23 @@
     from hsfs.validation_report import ValidationReport
 
 
+@public
```
manu-sj: I think this class can be left out to be private, the user ideally only access it through the `multi_part_insert` function

python/hsfs/feature_group_writer.py
```diff
@@ -28,13 +30,23 @@
...
+    materialization job only starts once, when the context exits.
+    """
+
     def __init__(self, feature_group):
         self._feature_group = feature_group
 
     def __enter__(self):
         return self
 
+    @public
```
manu-sj, line 48: I think this class can be left out to be private, the user ideally only access it through the multi_part_insert function

python/hopsworks_common/core/flink_cluster_api.py
robzor92: agreed

python/hopsworks_common/core/rest.py
robzor92: yes it is slop, ignoring for now

python/hopsworks_common/flink_cluster.py
robzor92: agreed

manu-sj (approved):

PR #1019, opened 2026-06-23
[HWORKS-1840] Upgrade KServe to v0.19.0

javierdlrm:
Part of the cross-repo KServe 0.19 upgrade (HWORKS-1840).
https://hopsworks.atlassian.net/browse/HWORKS-1840

Relax two dependency bounds so the KServe 0.19 Python client installs alongside hopsworks in the serving inference pipelines:

- `opensearch-py >=1.1.0,<=2.8.0` -> `>=2.8.0,<3.0.0`. opensearch-py 1.x caps urllib3<2, conflicting with KServe 0.19 (urllib3>=2.6). 2.8.0 allows urllib3>=2; the client code is already version-aware for 2.x.
- `protobuf >=4.25.4,<5.0.0` -> `>=4.25.4,<7.0.0`. KServe 0.19 requires protobuf>=6.33.5,<7. `uv lock` resolves to protobuf 6.33.6 with no conflict (uv.lock is gitignored here; pyproject.toml is the tracked source).

Companion PRs: hopsworks-helm #2035, docker-images #582, kserve #40 (merged).

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -50,10 +50,10 @@ dependencies = [
     "fsspec<2025.12.0",
     "retrying",
     "hopsworks_aiomysql[sa]==0.2.2",
-    "opensearch-py>=1.1.0,<=2.8.0",
+    "opensearch-py>=2.8.0,<3.0.0",   # >=2.8.0 for urllib3>=2 (KServe 0.19); <3.0.0 stays on the 2.x line
     "tqdm",
     "grpcio>=1.49.1,<2.0.0",         # ^1.49.1
-    "protobuf>=4.25.4,<5.0.0",       # ^4.25.4
+    "protobuf>=4.25.4,<7.0.0",       # widened to <7 so KServe 0.19 (protobuf>=6.33.5,<7) can coexist
     "packaging",                     # ^21.0
     "hopsworks-apigen>=1.0.4,<2.0.0",
     "click>=8.1",
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -50,7 +50,7 @@ dependencies = [
     "fsspec<2025.12.0",
     "retrying",
     "hopsworks_aiomysql[sa]==0.2.2",
-    "opensearch-py>=2.8.0,<3.0.0",   # >=2.8.0 for urllib3>=2 (KServe 0.19); <3.0.0 stays on the 2.x line
+    "opensearch-py>=1.1.0,<=2.8.0",
     "tqdm",
     "grpcio>=1.49.1,<2.0.0",         # ^1.49.1
     "protobuf>=4.25.4,<7.0.0",       # widened to <7 so KServe 0.19 (protobuf>=6.33.5,<7) can coexist


python/pyproject.toml
```diff
@@ -50,10 +50,10 @@ dependencies = [
     "fsspec<2025.12.0",
     "retrying",
     "hopsworks_aiomysql[sa]==0.2.2",
-    "opensearch-py>=1.1.0,<=2.8.0",
+    "opensearch-py>=2.8.0,<=2.8.0",  # 2.8.0 allows urllib3>=2 (KServe 0.19 needs urllib3>=2.6); 1.x capped urllib3<2
```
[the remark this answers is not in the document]
javierdlrm: Addressed: now `opensearch-py>=2.8.0,<3.0.0` (clear range, no awkward equal-bounds pin).

python/pyproject.toml
```diff
@@ -50,10 +50,10 @@ dependencies = [
     "fsspec<2025.12.0",
     "retrying",
     "hopsworks_aiomysql[sa]==0.2.2",
-    "opensearch-py>=1.1.0,<=2.8.0",
+    "opensearch-py==2.8.0",          # 2.8.0 allows urllib3>=2 (KServe 0.19 needs urllib3>=2.6); 1.x capped urllib3<2
```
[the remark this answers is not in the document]
javierdlrm: Addressed: relaxed to `>=2.8.0,<3.0.0`, so 2.8.x patch/security releases are allowed; `>=2.8.0` keeps urllib3>=2 (KServe 0.19) and `<3.0.0` excludes the 3.x major.

robzor92:
## Code review — KServe 0.14→0.19 (HWORKS-1840)

Clean, minimal 2-line dependency relaxation. **Mergeable as-is.** Full cross-repo context in hopsworks-helm #2035.

### ✅ Verified
- Widening `protobuf` to `<7.0.0` is safe: the SDK's legacy-style generated stub (`grpc_predict_v2_pb2.py`, no hard runtime version gate) loads, instantiates, and serializes correctly under both protobuf 4.25.4 and 6.33.6 — no codegen-vs-runtime break.
- The full union of SDK pins + KServe 0.19 pins (`opensearch-py>=2.8.0,<3` / `protobuf>=4.25.4,<7` / `grpcio>=1.49.1,<2` / `urllib3>=2.6`) resolves cleanly to urllib3 2.7.0, opensearch-py 2.8.0, protobuf 6.33.6, grpcio 1.81.1 (botocore, the historical `urllib3<2` capper, now allows `urllib3<3`).

### 🟢 Low / nits
- **Stale PR description** — it says `opensearch-py >=2.8.0,<=2.8.0`, but the diff/head is `>=2.8.0,<3.0.0`. Worth fixing so the description matches.
- **Blast radius of the opensearch-py floor.** Raising the lower bound 1.1.0→2.8.0 drops opensearch-py 1.x for **all** SDK installs, not just serving. Intentional (1.x caps `urllib3<2`), but flagging since the companion docker `agent-job` image still pins `opensearch-py==1.1.0` — that venv will be force-upgraded past its own pin once it installs the SDK. Worth reconciling the docker pin.

python/pyproject.toml
```diff
@@ -50,10 +50,10 @@ dependencies = [
     "fsspec<2025.12.0",
     "retrying",
     "hopsworks_aiomysql[sa]==0.2.2",
-    "opensearch-py>=1.1.0,<=2.8.0",
+    "opensearch-py>=2.8.0,<3.0.0",   # >=2.8.0 allows urllib3>=2 (KServe 0.19 needs urllib3>=2.6; 1.x capped urllib3<2); <3 avoids the 3.x major, 2.8.x patches allowed
```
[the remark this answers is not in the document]
javierdlrm: Fixed (ed97cbd9d): the inline comment is now accurate and concise — `<3.0.0` keeps the 2.x line (not just 2.8.x patches).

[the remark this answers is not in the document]
javierdlrm: Fixed (ed97cbd9d): comment reworded — `<3.0.0` stays on the 2.x line; the '2.8.x patches' wording is removed.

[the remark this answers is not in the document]
javierdlrm: Fixed: updated the PR description to the implemented constraint `>=2.8.0,<3.0.0`.

robzor92 (changes requested):
https://github.com/logicalclocks/hopsworks-api/tree/main/python/hopsworks_common/client/istio/grpc maybe needs to be refreshed?

javierdlrm:
> https://github.com/logicalclocks/hopsworks-api/tree/main/python/hopsworks_common/client/istio/grpc maybe needs to be refreshed?

not needed

```
▎ The client/istio/grpc stubs don't need refreshing for this upgrade:
▎ - protobuf 6 compat: the generated grpc_predict_v2_pb2.py imports cleanly under protobuf 6.33.6 (zero deprecation warnings) and message construct/serialize/parse round-trips. It uses the portable descriptor_pool.AddSerializedFile path, not the construction style that breaks on newer protobuf.
▎ - KServe 0.19 protocol: the vendored grpc_predict_v2.proto is identical (diff -w -B empty) to KServe 0.19's python/kserve/kserve/protocol/grpc/grpc_predict_v2.proto — same RPCs incl. RepositoryModelLoad/Unload, same fields. A refresh would only swap legacy→_builder codegen style (cosmetic).
```

robzor92 (approved):

python/pyproject.toml
```diff
@@ -50,10 +50,10 @@ dependencies = [
     "fsspec<2025.12.0",
     "retrying",
     "hopsworks_aiomysql[sa]==0.2.2",
-    "opensearch-py>=1.1.0,<=2.8.0",
+    "opensearch-py>=2.8.0,<3.0.0",   # >=2.8.0 for urllib3>=2 (KServe 0.19); <3.0.0 stays on the 2.x line
```
robzor92: **Recommend reverting this line.** It isn't needed for KServe, it breaks docker-images until that PR lands, and it's a breaking change for every SDK consumer.

The stated reason is urllib3>=2 for KServe 0.19. I isolated each bound with `uv pip compile` (py3.11) to check:

```
A: OLD opensearch (>=1.1.0,<=2.8.0) + OLD protobuf (<5) + kserve==0.19.0
   x No solution found:
     Because kserve==0.19.0 depends on protobuf>=6.33.5,<7.0.0 and you
     require protobuf>=4.25.4,<5.0.0, we can conclude that your requirements
     and kserve==0.19.0 are incompatible.

B: OLD opensearch (>=1.1.0,<=2.8.0) + NEW protobuf (<7) + kserve==0.19.0
   kserve==0.19.0
   opensearch-py==2.8.0      <-- 2.8.0 chosen anyway
   protobuf==6.33.6
   urllib3==2.7.0            <-- urllib3 2.x satisfied anyway
```

**Only the protobuf widening on `:56` is load-bearing.** Under the *old* bound the resolver already picks opensearch-py 2.8.0 and urllib3 2.7.0, because `<=2.8.0` made 2.8.0 the ceiling and resolvers prefer newest. Raising the floor buys nothing.

What it costs:

1. **It breaks docker-images.** `base-image/python-feature-pipeline/requirements.txt:14` and `base-image/agent-job/requirements.txt:14` pin `opensearch-py==1.1.0` and install it **after** the SDK, followed by a hard `pip check` under `set -e` → `hopsworks … requires opensearch-py>=2.8.0,<3.0.0, but you have opensearch-py 1.1.0` → image build fails. docker-images#582 bumps both, so this creates a **circular** dependency: #582 needs this PR's protobuf change, and this PR's opensearch floor needs #582. Dropping this line breaks the cycle and lets this PR merge first cleanly.
2. **It excludes users on older clusters.** `opensearch-py` is a *core* dependency (used only on feature-store vector-DB/search paths — `hopsworks_common/core/opensearch.py`, `hsfs/core/vector_db_client.py`). The platform ships OpenSearch **2.19.5** on the main line, where a 2.8 client is fine — but the **5.0 and 4.x release lines are still on OpenSearch 1.3.6** (`docker-images/opensearch/Jenkinsfile:13` on `branch-5.0`). opensearch-py 2.x is *documented* as supporting OpenSearch 1.x, but I had no cluster to confirm the SDK still works against a 5.0-line deployment, and that's the upgrade direction that matters.

If a floor is genuinely wanted somewhere, the narrow place is the image's own `requirements.txt`, not the SDK's public metadata.

Two side notes if you do keep it: the commit message says the bound became `>=2.8.0,<=2.8.0` but the code says `<3.0.0`; and the floor makes real code unreachable — `opensearch.py:108-111` `DEFAULT_OPTION_MAP`, the `< (2, 3)` branches at `:143`/`:154`, and the `f"{options[option]}s"` coercion at `:157`. Better deleted than left as dead branches. (The "already version-aware for 2.x" claim itself checks out — I verified on real 2.8.0 that `opensearchpy.__version__` is the tuple `(2, 8, 0)`, all four `opensearchpy.exceptions` imports resolve, and `get_default_py_config()`'s exact kwargs construct a `Urllib3HttpConnection` client on urllib3 2.7.0 with no warnings.)

python/pyproject.toml
```diff
@@ -50,10 +50,10 @@ dependencies = [
...
     "tqdm",
     "grpcio>=1.49.1,<2.0.0",         # ^1.49.1
-    "protobuf>=4.25.4,<5.0.0",       # ^4.25.4
+    "protobuf>=4.25.4,<7.0.0",       # widened to <7 so KServe 0.19 (protobuf>=6.33.5,<7) can coexist
```
robzor92, line 56: This line is the one that matters, and it's **verified safe** — see the review body for the round-trip results across 4.25.4 → 6.33.6. No change requested.

One correction for the commit message: it describes the vendored gencode as "protoc ~4.25". It's actually protoc 3.20-era old-style output (`_reflection` / `message_types_by_name` / `GeneratedProtocolMessageType`); KServe 0.19's own copy is 5.26.1 builder-style. The conclusion is right but the reason is the load-bearing part of the safety argument, so it's worth stating accurately — the reason it's safe is that pre-5.27 gencode contains no `ValidateProtobufRuntimeVersion` call for protobuf 6 to reject.

python/hopsworks_common/constants.py
```diff
@@ -39,6 +39,8 @@ class JOBS:
         "FRAMEWORK_FAILURE",
         "APP_MASTER_START_FAILED",
         "INITIALIZATION_FAILED",
+        "SUBMISSION_FAILED",
```
robzor92: **Keep the change, but the commit message's root cause is wrong, and this belongs in its own ticket.**

The `[append]` commit says QUEUE_TIMEOUT leaves `final_status` UNDEFINED, so `success` stays `None` and `_wait_until_finished` loops forever. The backend does the opposite:

- `jobs-ee/.../python/KubeJobsMonitor.java:379-384` — `updateFailedState()` calls `updateFinalStatus(JobFinalStatus.FAILED, execution)` at `:381`
- `hopsworks-common/.../ExecutionUpdateController.java:79-86` — persists it unconditionally, **no `isFinalState()` guard**

Both QUEUE_TIMEOUT producers (`KubeJobsMonitor.java:365`, `KubeDltIngestionMonitor.java:256`) go through `updateFailedState`, so the row is `state=QUEUE_TIMEOUT, final_status=FAILED`. Re-implementing `execution.py:175-190`:

| job | state | final_status | pre-fix | post-fix |
|---|---|---|---|---|
| PYTHON | QUEUE_TIMEOUT | FAILED | False | False | ← what the backend actually writes
| PYTHON | QUEUE_TIMEOUT | UNDEFINED | None | False | ← what the new test fabricates

Pre-fix, `execution.py:186` already returns `False` because `"FAILED"` is in `ERROR_STATES`. The described hang can't occur; the new test fails pre-fix *only* because it sets `final_status="UNDEFINED"`. (In fairness that matches the sibling PYTHON tests at `:367`/`:404` — but those pass pre-fix via `state`, so the convention hides the discrepancy here.)

**There is a real hazard, just a different one:** `updateState` and `updateFinalStatus` are two separate facade calls. If the monitor dies between them, `QUEUE_TIMEOUT`/`UNDEFINED` is permanent — and because `QUEUE_TIMEOUT` is absent from `getFinalStates()` (`JobState.java:155-159`), nothing ever reconciles it (`ExecutionFacade.findExpiredFinished` filters on that set). That's the accurate justification, and it's a good reason to make this change.

Two more things while you're in here:

- **`SUBMISSION_FAILED` is inert.** It's never persisted — Spark and Ray rewrite it to `FAILED` before saving (`KubeSparkJobsMonitor.java:376`, `RayJobsMonitor.java:390`). Harmless, but the message's claim that it "would cause the same hang" is also wrong.
- **`SUCCESS_STATES` (`:35`) omits `COMPLETED`**, while the server's `getFinalStates()` has 9 members. Currently unreachable (`COMPLETED` is normalized to `FINISHED` for Spark at `KubeSparkJobsMonitor.java:374` and for the python family at `KubeJobsMonitor.java:700-701`), so not a live bug — but it's the identical hang if any monitor ever persists it. Cheap to add defensively.

Finally, on scope: this has nothing to do with KServe. A reviewer auditing a dependency-bounds PR will not scrutinise `await_termination` semantics, which is exactly how the wrong-root-cause message would have slipped through.

python/tests/core/test_job.py
```diff
@@ -420,6 +420,45 @@ def test_run_await_termination_python_raises_on_killed(
         # Assert
         assert str(e_info.value) == "The Hopsworks Job was stopped"
 
+    def test_run_await_termination_python_raises_on_queue_timeout(
```
robzor92: This test only fails pre-fix because of the `final_status="UNDEFINED"` fixture on `:452` — see my note on `constants.py`. The backend actually writes `final_status=FAILED` for QUEUE_TIMEOUT, which `ERROR_STATES` already caught.

Worth adding a second case with `final_status="FAILED"` so the test pins the state the backend really produces, and (if you take the `COMPLETED` suggestion) one asserting the success path. The asserted message itself is correct — I traced it: for a non-YARN job `status = state` (`execution_engine.py:222`) and QUEUE_TIMEOUT misses the KILLED/FAILED/FRAMEWORK_FAILURE branches, so `:236-238` produces exactly that string.

Separate follow-up, not for this PR: the fix converts "hang forever" into "hang 6 minutes". Once `success is False`, `execution_engine.py:196-216` still waits for stdout/stderr log aggregation with `await_time = 6*60`, and `job.run(await_termination=True)` passes `timeout=None` (`job.py:410`), so the `if timeout and …` guards at `:203`/`:213` never fire → 120 polls × 3s = **360s before raising**. A queue-timed-out job never produced logs, so that wait is pure latency.

robzor92 (changes requested):
## Cross-repo review (part of HWORKS-1840: helm#2035, docker-images#582, hopsworks-ee#3122, tools#258)

Reviewing this as one change across the five PRs. Short version: **the protobuf widening is load-bearing and I verified it's safe. The opensearch-py change is neither needed nor free, and it creates a circular merge dependency with docker-images#582.** The `[append]` commit is fine as a change but its stated root cause is wrong.

### The protobuf widening is verified safe — no action needed

I checked the thing that would actually break here, since this is a public PyPI package and a fresh `pip install hopsworks` is the blast radius. The SDK *does* vendor gencode (`python/hopsworks_common/client/istio/grpc/proto/grpc_predict_v2_pb2.py`), but it's pre-4.21 old-style output (`_reflection`, `message_types_by_name`, `GeneratedProtocolMessageType`) with no `runtime_version.ValidateProtobufRuntimeVersion` call to trip — that check only exists in gencode from protoc 5.27+. Import plus a deterministic round-trip across the whole declared range gave byte-identical wire output:

```
protobuf=4.25.4  bytes=138  sha256=3628ac9f…  roundtrip equal: True
protobuf=5.29.5  bytes=138  sha256=3628ac9f…  roundtrip equal: True
protobuf=6.33.5  bytes=138  sha256=3628ac9f…  roundtrip equal: True
protobuf=6.33.6  bytes=138  sha256=3628ac9f…  roundtrip equal: True
```

No deprecation warnings on 6.33.6. No wire risk either — the vendored serialized `FileDescriptorProto` is byte-identical to `kserve==0.19.0`'s (3445 bytes, zero message/field/RPC diff), and importing both in one process dedupes to the same `DESCRIPTOR` under upb rather than colliding in the pool. Good news given `uv.lock` is gitignored, so `uv sync` re-resolves on every CI run and a resolution failure would surface.

### CI gap worth knowing

Nothing in this repo installs `kserve` alongside `hopsworks`, so the co-existence this PR exists for is never exercised here. `python.yml` does genuinely exercise the vendored gencode on protobuf 6 (the `dev` extra resolves `protobuf==6.33.6` across 3.10–3.13), which covers the risk above — but the kserve pairing is only ever tested in docker-images, and only if that PR's `bake-base` label is added.

### Minor
No changelog/release-notes file exists in the repo. A dependency *floor* raise on a public package normally warrants one — a user on opensearch-py 1.x gets a bare resolver error from `pip install -U hopsworks` with no documented cause. Moot if the floor change is dropped per my inline comment.

python/pyproject.toml
javierdlrm: Reverted (042ba539f): back to `>=1.1.0,<=2.8.0`; only the protobuf widening remains. This also breaks the circular ordering with docker-images#582, so this PR can merge first.

python/pyproject.toml
javierdlrm: Kept unchanged. Correction acknowledged: the load-bearing safety reason is that the vendored gencode is pre-5.27 (protoc 3.20-era output), so it contains no ValidateProtobufRuntimeVersion call for protobuf 6 to reject.

python/hopsworks_common/constants.py
javierdlrm: Acknowledged, and keeping the change with your corrected justification: the real hazard is the monitor dying between the separate updateState and updateFinalStatus facade calls, with QUEUE_TIMEOUT absent from getFinalStates so nothing ever reconciles the row. Splitting this (plus the SUCCESS_STATES/COMPLETED hardening) into its own ticket with an accurate message is on the table before merge.

python/tests/core/test_job.py
javierdlrm: Agreed: the `final_status="FAILED"` fixture is the one that pins what the backend actually writes. It will go in together with however the job-states change is split out (see the constants.py thread).

python/hopsworks_common/constants.py
javierdlrm: Done: the change is out of this PR (dropped in the 16ab94b59 force-push; this diff is protobuf-only now) and lives in its own ticket and PR with the corrected root cause: HWORKS-2981 / #1074. The commit message there attributes the hang to the monitor-crash window between updateState and updateFinalStatus (QUEUE_TIMEOUT missing from getFinalStates, so findExpiredFinished never reconciles), describes SUBMISSION_FAILED and the new COMPLETED addition to SUCCESS_STATES as defensive, and drops the wrong claims.

python/tests/core/test_job.py
javierdlrm: Done in #1074: a second case with `final_status="FAILED"` pins the state the backend actually writes (green pre-fix via ERROR_STATES, guarding the production path), the `UNDEFINED` case stays as the stranded-row hang guard with its comment corrected to the crash-window rationale, and `COMPLETED` joined SUCCESS_STATES defensively.

javierdlrm:
## Review follow-up, round 2

This PR is now **protobuf-only**: the QUEUE_TIMEOUT/job-states change was dropped (16ab94b59 force-push) and split into HWORKS-2981 / #1074 with the corrected root-cause message, the `final_status="FAILED"` fixture pinning what the backend actually writes, and defensive `COMPLETED` in SUCCESS_STATES. With the opensearch floor already reverted, this PR carries exactly the one load-bearing line and can merge first.

robzor92 (approved):

PR #1038, opened 2026-07-02
[FSTORE-2055] incremental training datasets

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```markdown
```


python/hsfs/core/transformation_function_engine.py
```diff
@@ -1272,6 +1272,12 @@ def _fit_and_transform(
         Returns:
             The transformed dataset, in the same shape as `dataset`.
         """
+        # A standalone training dataset (created directly, not from a feature
```
manu-sj, line 1275: I though this behaviour was deprecated. I think all training datasets should have feature views now.

I don't think there is any issue with this check it can be left for safety but was just wondering if this had to be added because of some issue encountered.

utils/python/hsfs_utils.py
```diff
@@ -101,8 +101,11 @@ def create_td(job_conf: dict[Any, Any]) -> None:
     feature_store = job_conf.pop("feature_store")
     fs = get_feature_store_handle(feature_store)
 
-    # Extract the query object
-    q = query.Query._hopsworks_json(job_conf.pop("query"))
+    # Extract the query object. Use the full deserializer (not _hopsworks_json)
```
manu-sj, line 104: Just curious is this code path used. I thought `create_td` was legacy code and was not used and It was `create_fv_td` that was used everywhere

python/hsfs/core/transformation_function_engine.py
bubriks: This was added since i tried to create/append data to TD using FV and in stand alone approach it was failing here.

I can undo it, but then maybe we should deprecate/remove `get_training_dataset` and `create_training_dataset` from `feature_store.py`. If i remember correctly if TD was created using FV but received using `fs.get_training_dataset` it also fails since it doesnt have FV reference.

python/hsfs/core/feature_view_engine.py
```diff
@@ -903,6 +985,13 @@ def _compute_training_dataset(
             training_dataset_version=training_dataset_obj.version,
         )
 
+        # On append the newly written partition is a small daily increment, but
```
manu-sj, line 1065: Not recomputing training data statistics after appends is a correctness bug. If a user adds training data that changes the statistical properties of the split, such as min, mean, or max, then transformation functions that rely on those statistics may transform the data incorrectly.

Some statistics can likely be updated incrementally. For example, max can be computed as `max(previous_training_data_max, new_training_data_max)`. However, I am not sure whether this is possible for all of the statistics we currently support, or how much work it would require.

I agree that recomputing the statistics automatically here could be extremely inefficient. However, I think we should at least provide users with a way to explicitly trigger a recomputation of statistics for a training dataset version after an insert.

And this complexity is only required if the feature view has transformation functions that use statistics.

python/hsfs/feature_view.py
```diff
@@ -2606,6 +2606,163 @@ def recreate_training_dataset(
...
+
+        With `overwrite=True` the entire training dataset version is rewritten instead, equivalent to [`recreate_training_dataset`][hsfs.feature_view.FeatureView.recreate_training_dataset] for the given window.
+
+        Statistics are not recomputed on append by default, since that reads the whole (potentially multi-terabyte) dataset back on every append.
+        Pass `compute_statistics=True` to refresh the descriptive statistics over all increments after the batch is written, or call [`compute_training_dataset_statistics`][hsfs.feature_view.FeatureView.compute_training_dataset_statistics] explicitly when needed, e.g. periodically or right before retraining.
+        Model-dependent transformation functions transform each appended batch with the statistics computed when the version was created, so all increments and serving stay consistent with each other; to refit those statistics, rebuild the version with `overwrite=True` or create a new version.
```
manu-sj: > Model-dependent transformation functions transform each appended batch with the statistics computed when the version was created, so all increments and serving stay consistent with each other; to refit those statistics, rebuild the version with `overwrite=True` or create a new version.

Wouldn't `compute_statistics=True` also fix this problem?

python/hsfs/feature_view.py
```diff
@@ -2606,6 +2606,163 @@ def recreate_training_dataset(
...
+
+        Randomly split training datasets are appended per split: each split receives a share of the batch (e.g. 80/10/10), which is sound over many appends.
+        Appending to a time-series-split training dataset is not supported and raises an error: the batch would land entirely in the last split (e.g. test) while the earlier splits stay frozen, skewing the dataset with every append.
```
manu-sj: I think we can also remove this line : _"the batch would land entirely in the last split (e.g. test) while the earlier splits stay frozen, skewing the dataset with every append"_ Since we already mentioned that it is not supported

python/hsfs/feature_view.py
```diff
@@ -2606,6 +2606,163 @@ def recreate_training_dataset(
...
+            )
+            X_test, y_test = feature_view.get_training_data(
+                training_dataset_version=1,
+                start_time="2026-06-02",
+            )
+            ```
+
+        Warning: Engine Support
+            Appending (`overwrite=False`) is only supported for the `parquet` data format.
+            On the `python` engine the append is offloaded to the Hopsworks Feature Query Service (FlyingDuck) or a backend Spark job; it requires a recent enough Query Service, as older versions overwrite instead of appending.
```
manu-sj: > it requires a recent enough Query Service, as older versions overwrite instead of appending.

Maybe we can mention the version of Hopsworks from which this feature is available instead. Atleast for now since flyingduck version is related to the Hopsworks version that would make more sense for users I think.

python/hsfs/core/feature_view_engine.py
```diff
@@ -706,6 +729,109 @@ def _recreate_training_dataset(
...
+        # time-series dataset is an unsplit one read with a time range.
+        if not overwrite and any(
+            split.split_type == TrainingDatasetSplit.TIME_SERIES_SPLIT
+            for split in training_dataset_obj.splits
+        ):
+            raise FeatureStoreException(
+                "Appending to a time-series-split training dataset is not "
+                "supported: the new batch would land entirely in the last "
+                "split (e.g. test) while the earlier splits stay unchanged, "
+                "skewing the dataset. For a growing time-series training "
```
manu-sj: I think we can remove this part  in the error message, mostly because this is in internal implementation detail that can be fixed to some extended by 

> : the new batch would land entirely in the last "
>                 "split (e.g. test) while the earlier splits stay unchanged, "
>                 "skewing the dataset

python/hsfs/feature_view.py
```diff
@@ -2606,6 +2606,163 @@ def recreate_training_dataset(
...
+        self,
+        training_dataset_version: int,
+        start_time: str | int | datetime | date | None = "",
+        end_time: str | int | datetime | date | None = "",
+        overwrite: bool = False,
+        write_options: dict[Any, Any] | None = None,
+        spine: SplineDataFrameTypes | None = None,
+        transformation_context: dict[str, Any] | None = None,
+        compute_statistics: bool = False,
+        partition_precision: str = "day",
```
manu-sj: What happens if an user changes the partition precision when writing the training data. Would that corrupt the partisions?

Maybe `partition_precision` should be a parameters of the training_dataset creations functions and stored with the training dataset metatdata. Because I don't think we should be able to change the partition once we created the initial training data.

python/hsfs/core/feature_view_engine.py
```diff
@@ -706,6 +729,109 @@ def _recreate_training_dataset(
...
+        # full-dataset read here. On overwrite the statistics were already
+        # recomputed by the materialization itself.
+        if compute_statistics and not overwrite:
+            self._recompute_training_dataset_statistics(
+                feature_view_obj, training_dataset_version
+            )
+        return training_dataset_obj, td_job
+
+    @staticmethod
+    def _event_date_int(timestamp_ms):
```
manu-sj: We have a lot of datetime related convertion functions here https://github.com/logicalclocks/hopsworks-api/blob/c01a04788557bcbcae8ad3d9828fdeb117002690/python/hopsworks_common/util.py#L278 

Maybe we move this one of there, but leaving it here is also fine

python/hsfs/core/transformation_function_engine.py
manu-sj: I agree I think we should remove it, but maybe we do in a different PR

manu-sj (approved):
I was looking into the changes done for this PR's I think it works very well for random splits but I think we might need some more improvement for time series split.

PR #1042, opened 2026-07-06
[HWORKS-2879][HWORKS-3145] Spark upgrade to 4.1. Hudi upgrade to 1.2.0

gibchikafa:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```markdown
```


.github/workflows/java.yml
```diff
@@ -29,7 +29,7 @@ jobs:
             ${{ runner.os }}-maven-
       - name: Test
         working-directory: ./java
-        run: mvn clean test -Pspark-3.5 -Dspark.version=3.5.5 -Dhudi.version=1.0.2
+        run: mvn clean test -Pspark-4.1 -Dspark.version=4.1.2 -Dhudi.version=1.2.0
```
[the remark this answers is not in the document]
gibchikafa: The version difference is intentional: the `spark-4.1` profile pins the Hops Spark fork (`4.1.2.0`), which CI runners cannot resolve from public repositories, so CI overrides it to build against upstream Apache Spark `4.1.2`. Aligning CI to `4.1.2.0` was tried and reverted (see "Align Spark versions in CI" / "Restore CI-compatible Spark pins"). Added a comment in the workflow documenting this, and dropped the redundant `-Dhudi.version` flag so the profile controls the Hudi version. See 3e97862fd.

.github/workflows/java.yml
```diff
@@ -58,9 +58,9 @@ jobs:
 
       - name: Test
         working-directory: ./java
-        run: mvn clean test -Pspark-3.5 -Dspark.version=3.5.5 -Dhudi.version=1.0.2
+        run: mvn clean test -Pspark-4.1 -Dspark.version=4.1.2 -Dhudi.version=1.2.0
```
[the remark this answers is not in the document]
gibchikafa: The version difference is intentional: the `spark-4.1` profile pins the Hops Spark fork (`4.1.2.0`), which CI runners cannot resolve from public repositories, so CI overrides it to build against upstream Apache Spark `4.1.2`. Aligning CI to `4.1.2.0` was tried and reverted (see "Align Spark versions in CI" / "Restore CI-compatible Spark pins"). Added a comment in the workflow documenting this, and dropped the redundant `-Dhudi.version` flag so the profile controls the Hudi version. See 3e97862fd.

.github/workflows/java.yml
```diff
@@ -88,4 +88,4 @@ jobs:
...
```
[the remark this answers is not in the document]
gibchikafa: The version difference is intentional: the `spark-4.1` profile pins the Hops Spark fork (`4.1.2.0`), which CI runners cannot resolve from public repositories, so CI overrides it to build against upstream Apache Spark `4.1.2`. Aligning CI to `4.1.2.0` was tried and reverted (see "Align Spark versions in CI" / "Restore CI-compatible Spark pins"). Added a comment in the workflow documenting this, and dropped the redundant `-Dhudi.version` flag so the profile controls the Hudi version. See 3e97862fd.

.github/workflows/mkdocs-test.yml
```diff
@@ -52,15 +52,15 @@ jobs:
...
         with:
-          java-version: "8"
-          distribution: "adopt"
+          java-version: "17"
+          distribution: "temurin"
 
       - name: Build javadoc documentation
         working-directory: java
-        run: mvn clean install javadoc:javadoc javadoc:aggregate -DskipTests -Dspark.version=3.5.5 -Dhudi.version=1.0.2 && cp -r target/site/apidocs ../logicalclocks.github.io/docs/javadoc
+        run: mvn clean install javadoc:javadoc javadoc:aggregate -Pspark-4.1 -DskipTests -Dspark.version=4.1.2 -Dhudi.version=1.2.0 && cp -r target/site/apidocs ../logicalclocks.github.io/docs/javadoc
```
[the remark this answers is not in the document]
gibchikafa: The version difference is intentional: the `spark-4.1` profile pins the Hops Spark fork (`4.1.2.0`), which CI runners cannot resolve from public repositories, so CI overrides it to build against upstream Apache Spark `4.1.2`. Aligning CI to `4.1.2.0` was tried and reverted (see "Align Spark versions in CI" / "Restore CI-compatible Spark pins"). Added a comment in the workflow documenting this, and dropped the redundant `-Dhudi.version` flag so the profile controls the Hudi version. See 3e97862fd.

.github/workflows/java.yml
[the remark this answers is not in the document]
gibchikafa: Clarification after digging into the failed run for 2b3ec9b0c: the fork's Spark artifacts are in fact public (nexus.hops.works/repository/spark), but their poms declare dependencies on `io.hops:hadoop-client-*:3.4.3.2-EE-RC0` and `io.hops.hive:*:4.1.0`, which live only in the authenticated hops-artifacts repository (401 anonymously). Maven fails while reading those artifact descriptors, so CI — and fork PRs especially, which cannot use repo secrets — has to build against upstream Apache Spark. The workflow comments now state this precisely.

.github/workflows/java.yml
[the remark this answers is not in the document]
gibchikafa: Clarification after digging into the failed run for 2b3ec9b0c: the fork's Spark artifacts are in fact public (nexus.hops.works/repository/spark), but their poms declare dependencies on `io.hops:hadoop-client-*:3.4.3.2-EE-RC0` and `io.hops.hive:*:4.1.0`, which live only in the authenticated hops-artifacts repository (401 anonymously). Maven fails while reading those artifact descriptors, so CI — and fork PRs especially, which cannot use repo secrets — has to build against upstream Apache Spark. The workflow comments now state this precisely.

.github/workflows/java.yml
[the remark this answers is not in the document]
gibchikafa: Clarification after digging into the failed run for 2b3ec9b0c: the fork's Spark artifacts are in fact public (nexus.hops.works/repository/spark), but their poms declare dependencies on `io.hops:hadoop-client-*:3.4.3.2-EE-RC0` and `io.hops.hive:*:4.1.0`, which live only in the authenticated hops-artifacts repository (401 anonymously). Maven fails while reading those artifact descriptors, so CI — and fork PRs especially, which cannot use repo secrets — has to build against upstream Apache Spark. The workflow comments now state this precisely.

.github/workflows/mkdocs-test.yml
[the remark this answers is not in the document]
gibchikafa: Clarification after digging into the failed run for 2b3ec9b0c: the fork's Spark artifacts are in fact public (nexus.hops.works/repository/spark), but their poms declare dependencies on `io.hops:hadoop-client-*:3.4.3.2-EE-RC0` and `io.hops.hive:*:4.1.0`, which live only in the authenticated hops-artifacts repository (401 anonymously). Maven fails while reading those artifact descriptors, so CI — and fork PRs especially, which cannot use repo secrets — has to build against upstream Apache Spark. The workflow comments now state this precisely.

java/pom.xml
```diff
@@ -308,6 +308,23 @@
                 <kafka.version>3.4.1</kafka.version>
             </properties>
         </profile>
+        <profile>
+            <id>spark-4.1</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
```
robzor92, line 308: `activeByDefault` is dropped by Maven as soon as any other profile activates — so `mvn deploy -Pwith-hops-ee` (per the comment below) silently deactivates spark-4.1 and falls back to the internally inconsistent top-level properties (spark 4.1.3.0 + scala 2.12 + hudi 1.0.2.1 → unresolvable `spark-core_2.12:4.1.3.0`). Release/EE pipelines must add `-Pspark-4.1`; CONTRIBUTING.md still documents `-Pspark-3.5,with-hops-ee`.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/util/StorageConnectorUtils.java
```diff
@@ -332,7 +331,7 @@ public Dataset<Row> read(StorageConnector connector, DataSource dataSource, Stri
     } else if (connector instanceof StorageConnector.GlueConnector) {
       return read((StorageConnector.GlueConnector) connector, dataSource, dataFormat, options);
     } else if (connector instanceof StorageConnector.KafkaConnector) {
-      throw new NotSupportedException("Reading a Kafka Stream into a static Spark Dataframe is not supported.");
+      throw new UnsupportedOperationException("Reading a Kafka Stream into a static Spark DataFrame is not supported.");
```
robzor92, line 334: Type change from `javax.ws.rs.NotSupportedException` — both unchecked, but callers catching the old type break. Intentional?

robzor92 (commented):
The Spark 4 / Scala 2.13 adaptations look right (`partitionBy` Seq→array conversion, `sourceOrderingFields` rename, the JDK17 doclint/lombok work), CI is green across the matrix, and the python side needed pleasantly little. My comments are mostly about build/packaging and cross-repo consistency.

**Build/packaging**

- The `spark-4.1` profile is `activeByDefault` (inline): Maven deactivates it when any other profile is activated, so `mvn deploy -Pwith-hops-ee` alone (which the in-pom comment says is required for deploy) falls back to spark 4.1.3.0 + scala 2.12 + hudi 1.0.2.1 → unresolvable `spark-core_2.12:4.1.3.0`. Fails loudly, but every release/EE pipeline must now add `-Pspark-4.1`. CONTRIBUTING.md line ~323 still documents `-Pspark-3.5,with-hops-ee`.
- The artifact rename `hsfs-spark-spark3.5` → `hsfs-spark-spark4.1` is a published interface: docker-images jar URLs, ee, and chart job configs need a consumer sweep before merge. docker-images#872 currently pulls a `-jar-with-dependencies` jar from `/dev/gibson/` — the final URL/naming must match what the release job actually publishes for the new artifactId.
- GroupId question: this compiles against `org.apache.hudi:*:1.2.0` (provided) while the fork (hudi#33) publishes `io.hops.hudi:*:1.2.0` and the runtime bundle comes from repo.hops.works. Works if the fork's API surface is identical — worth confirming the compile-vs-runtime coordinate divergence is intended.

**Minor**

- `java/spark/pom.xml` (with-hops-ee) still pins `io.hops:hadoop-client:3.4.3.1-EE-RC0` while utils bumped to 3.4.3.2-EE-RC0 and the Spark fork builds against 3.4.3.2 — align.
- Workflow ivy cache keys still say `ivy-spark-avro-3.5.5` (coverage.yml + python.yml ×8) while fetching `spark-avro_2.13:4.1.1` — the cache never hits, jars re-download every run.
- `StorageConnectorUtils` (inline): `NotSupportedException` → `UnsupportedOperationException` changes the catchable type for callers.
- `deequ.version` is declared in both profiles but consumed by nothing (profiling moved to datasketches) — dead property, drop it.
- The pandas<2.2 skip guards mean the "Unit Tests (Pandas 1.x)" job now skips all Spark tests — intentional coverage reduction, fine, just noting.

**Note:** Spark 4 defaults `spark.sql.ansi.enabled=true` and nothing in the epic sets it — engine cast paths (e.g. python/hsfs/engine/spark.py casts, partition_grains) rely on lenient cast semantics. Needs a platform-level stance (spark-defaults in the chart, or a release note); flagged on the helm PR too.

java/pom.xml
gibchikafa: Fixed in 5bca42325: the top-level properties now mirror the `spark-4.1` profile exactly (scala 2.13.17, `org.apache.hudi` 1.2.0, avro 1.12.1, jackson 2.21.2, `artifact.spark.version=spark4.1`), so a build where Maven drops the activeByDefault profile — `mvn deploy -Pwith-hops-ee` included — resolves the identical stack. Verified with `mvn help:evaluate`: the default build, `-Pwith-hops-ee`, and `-Pspark-4.1,with-hops-ee` all produce the same spark/scala/hudi/artifact values. CONTRIBUTING.md now documents `-Pspark-4.1,with-hops-ee` for explicitness anyway. Same commit also drops the dead `deequ.version` properties and aligns the with-hops-ee `hadoop-client` pin to 3.4.3.2-EE-RC0.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/util/StorageConnectorUtils.java
gibchikafa: Intentional, and effectively forced: `javax.ws.rs.NotSupportedException` was only on the classpath transitively via the Spark 3.5/Hudi tree, and nothing in the Spark 4.1/Hudi 1.2 tree provides `javax.ws.rs` anymore (`grep -r javax.ws.rs java/` is empty — there was never a direct dependency). It was also semantically a misuse — that's JAX-RS's HTTP 415 'Unsupported Media Type' server exception. Callers catching the old type would break, but they'd break at class-load anyway once the transitive jar disappeared; `UnsupportedOperationException` is the standard JDK type for this.

java/pom.xml
```diff
@@ -28,13 +28,13 @@
...
+        <spark.version>4.1.2.0</spark.version>
+        <hudi.groupId>io.hops.hudi</hudi.groupId>
         <hudi.version>1.0.2.1</hudi.version>
         <awssdk.version>2.10.40</awssdk.version>
         <scala.version>2.12.10</scala.version>
         <scala-short.version>2.12</scala-short.version>
-        <dbutils.version>0.0.5</dbutils.version>
-        <json.version>20231013</json.version>
+<json.version>20231013</json.version>
         <hoverfly.version>0.12.2</hoverfly.version>
```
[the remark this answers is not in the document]
gibchikafa: Fixed in 3e97862fd.

utils/java/pom.xml
```diff
@@ -73,7 +73,11 @@
             <exclusions>
                 <exclusion>
                     <groupId>io.hops.hudi</groupId>
-                    <artifactId>hudi-spark-common_2.12</artifactId>
+                    <artifactId>hudi-spark-common_2.13</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>io.hops.hudi</groupId>
+                    <artifactId>hudi-spark_2.12</artifactId>
                 </exclusion>
```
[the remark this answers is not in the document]
gibchikafa: Fixed in 3e97862fd.

utils/java/pom.xml
```diff
@@ -84,7 +88,7 @@
              they're transitively present in hsfs-spark-spark3.5. -->
```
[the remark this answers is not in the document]
gibchikafa: Fixed in 3e97862fd.

utils/java/pom.xml
```diff
@@ -60,10 +60,10 @@
              spark-feature-pipeline/spark_install.sh). -->
         <dependency>
             <groupId>io.hops.hudi</groupId>
-            <artifactId>hudi-utilities_2.12</artifactId>
+            <artifactId>hudi-utilities_2.13</artifactId>
             <version>${hudi.version}</version>
             <scope>provided</scope>
-            <!-- hudi-spark-common_2.12:1.0.2.1 is published corrupt on archiva
+            <!-- hudi-spark-common_2.13:1.0.2.1 is published corrupt on archiva
```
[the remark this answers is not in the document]
gibchikafa: Fixed in 3e97862fd.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/util/StorageConnectorUtils.java
```diff
@@ -332,7 +331,7 @@ public Dataset<Row> read(StorageConnector connector, DataSource dataSource, Stri
     } else if (connector instanceof StorageConnector.GlueConnector) {
       return read((StorageConnector.GlueConnector) connector, dataSource, dataFormat, options);
     } else if (connector instanceof StorageConnector.KafkaConnector) {
-      throw new NotSupportedException("Reading a Kafka Stream into a static Spark Dataframe is not supported.");
+      throw new UnsupportedOperationException("Reading a Kafka Stream into a static Spark Dataframe is not supported.");
     } else {
```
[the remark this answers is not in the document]
gibchikafa: Fixed in 3e97862fd.

gibchikafa:
**Fixed**
- Top-level `java/pom.xml` properties now mirror the `spark-4.1` profile exactly, so a build where Maven drops the activeByDefault profile (`mvn deploy -Pwith-hops-ee` alone) resolves the identical stack — verified with `mvn help:evaluate` for the default build, `-Pwith-hops-ee`, and `-Pspark-4.1,with-hops-ee`. CONTRIBUTING.md now documents `-Pspark-4.1,with-hops-ee`.
- `with-hops-ee` `hadoop-client` pin aligned to 3.4.3.2-EE-RC0.
- Ivy cache keys renamed `ivy-spark-avro-3.5.5` → `ivy-spark-avro-4.1.1` (coverage.yml + python.yml ×8) so the cache actually hits.
- Dead `deequ.version` properties dropped from both profiles.

**Acknowledged, tracked outside this PR**
- Artifact rename consumer sweep: agreed `hsfs-spark-spark4.1` must be swept through docker-images/ee/chart before merge; docker-images#872 will be repointed from `/dev/gibson/` to the URL the release job actually publishes for the new artifactId.
- ANSI mode: agreed `spark.sql.ansi.enabled=true` needs a platform-level stance rather than per-job settings; following up on the helm PR.
- Pandas 1.x job skipping all Spark tests is intentional: PySpark 4.1 requires pandas ≥2.2, so that job keeps only non-Spark coverage.

gibchikafa:
Update on the Hudi groupId point: rather than compiling against upstream, 3857e415b switches `hudi.groupId` to the fork (`io.hops.hudi`) in both the top-level properties and the `spark-4.1` profile, so compile and runtime coordinates now agree (verified `io.hops.hudi:hudi-spark4.1-bundle_2.13:1.2.0` and `hudi-utilities-slim-bundle_2.13:1.2.0` resolve from hops-artifacts). The fork will additionally be published to an open repository (nexus.hops.works/repository/hudi), which the pom now declares; until it is populated, CI temporarily overrides `-Dhudi.groupId=org.apache.hudi`, mirroring the existing `spark.version` override, and that flag can be dropped once the open repo has the artifacts.

robzor92 (approved):

java/spark/src/main/java/com/logicalclocks/hsfs/spark/util/StorageConnectorUtils.java
robzor92: **Checked the reasoning — it holds. No change needed.**

`javax.ws.rs` appears nowhere in `java/` at head (268 files), and `StorageConnectorUtils.java:334` now throws `UnsupportedOperationException`. Since `javax.ws.rs` only ever arrived transitively through the Spark 3.5 / Hudi tree and nothing in the Spark 4.1 / Hudi 1.2 tree provides it, a caller catching the old type would hit a `NoClassDefFoundError` at class-load regardless of what we throw — so there is no compatibility left to preserve.

`UnsupportedOperationException` is also the semantically correct type; the JAX-RS one is an HTTP 415 "Unsupported Media Type" server exception.

java/pom.xml
```diff
@@ -318,6 +318,22 @@
                 <kafka.version>3.4.1</kafka.version>
             </properties>
         </profile>
+        <profile>
+            <id>spark-4.1</id>
```
o-alex, line 306: **Must Fix.** The fix for the `activeByDefault` problem moved the inconsistency rather than removing it.

Now that the top-level properties mirror `spark-4.1` exactly, the older profiles only override a subset. `spark-3.5` sets `spark.version`, `deequ.version`, `artifact.spark.version`, `fasterxml.jackson.databind.version` and `kafka.version`, but **not** `scala.version`, `scala-short.version`, `hudi.groupId` or `hudi.version`. So `-Pspark-3.5` now inherits scala 2.13.17 and `io.hops.hudi:1.2.0` from the top level and resolves:

- `spark-core_2.13:3.5.5.2` (the Hops 3.5 fork is 2.12, so this does not exist)
- `hudi-spark3.5-bundle_2.13:1.2.0` (does not exist)

`spark-3.3` is worse, it only sets `artifact.spark.version`. Before this change the top-level defaults *were* the 3.5 stack, so those profiles worked.

Two clean options: delete the `spark-3.3` and `spark-3.5` profiles if 5.1 no longer builds them, or give them the full property set. Leaving them half-populated means the next person who runs `-Pspark-3.5` gets an unresolvable build.

python/hsfs/core/iceberg_engine.py
```diff
@@ -1030,11 +1030,15 @@ def _snapshot_read_opts_at(self, location: str, end_ts: int) -> dict[str, str]:
         """
         snapshots = self._read_snapshots(location)
         if not snapshots:
-            return {self.ICEBERG_QUERY_TIME_TRAVEL_AS_OF_TIMESTAMP: str(end_ts)}
+            return {
+                self.ICEBERG_QUERY_TIME_TRAVEL_AS_OF_INSTANT: util._get_delta_datestr_from_timestamp(
```
o-alex, line 1034: **Should Fix (verify on a cluster).** This reuses the **Delta** timestamp formatter for an Iceberg read option. `_get_delta_datestr_from_timestamp` produces `2026-08-28 10:15:30.123Z`: a space separator with a trailing `Z`, which is not ISO-8601.

The old code passed epoch millis to `as-of-timestamp`, where the format was unambiguous. Delta's `timestampAsOf` and Iceberg's `timestampAsOf` are not documented to accept the same string shapes, and an unrecognised read option does not raise, Spark just ignores it. So if the format is wrong the failure mode is "time travel silently returns the latest snapshot", not a stack trace.

The updated test only asserts dict equality against a mock, so it pins the formatting but proves nothing about whether Iceberg accepts it. This is exactly what loadtest#997's Iceberg PIT coverage should catch, so it would be good not to merge these two independently of each other.

python/hsfs/core/iceberg_engine.py
```diff
@@ -1012,7 +1012,7 @@ def _snapshot_read_opts_at(self, location: str, end_ts: int) -> dict[str, str]:
         The snapshot is resolved client-side against the snapshots' own
         ``committed_at`` clock, the clock the backend derives the recorded
         commit time from.
-        Passing ``as-of-timestamp`` instead lets Iceberg resolve the bound
+        Passing ``timestampAsOf`` instead lets Iceberg resolve the bound
```
o-alex, line 1015: **Should Fix (docs).** The constant rename was applied to the prose as well as the code, and it inverted the meaning. This paragraph now argues *against* passing `timestampAsOf`, explaining that it resolves server-side against a different clock and can silently read snapshot N-1, while the `if not snapshots:` branch a few lines below does exactly that.

Before the rename the sentence was about `as-of-timestamp` and was correct. It needs rewriting rather than renaming.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/FeatureGroup.java
```diff
@@ -61,6 +61,11 @@ public class FeatureGroup extends FeatureGroupBase<Dataset<Row>> {
   protected FeatureGroupEngine featureGroupEngine = new FeatureGroupEngine();
   protected StatisticsEngine statisticsEngine = new StatisticsEngine(EntityEndpointType.FEATURE_GROUP);
 
+  public static class FeatureGroupBuilder {
```
o-alex, line 64: **Should Fix (small).** This empty builder shim appears four times across the PR (`FeatureGroup`, `ExternalFeatureGroup`, `TrainingDataset`, and `StreamFeatureGroup` in hsfs/beam/flink) with no explanation.

Assuming it is there so lombok's generated builder class carries a javadoc-able declaration under JDK 17's stricter doclint, one line saying so on one of them would save the next reader working out why four apparently pointless classes exist.

o-alex (commented):
Joint review by Claude Code and an independent Codex pass, against `9081f88a`. Noting this PR is already approved, so the first item below is worth a look before it merges.

Checked and clear: the Hudi `EIGHT` to `NINE` table-version migration logic is structurally identical to what it replaces and handles the important case (a production table at v8 migrates straight to v9); `partitionBy(String[])` compiles because Spark's `DataFrameWriter.partitionBy` is `@scala.annotation.varargs`; the coordinate wiring across the three poms is self-consistent for the default and `-Pspark-4.1` builds.

1 must-fix, 3 should-fix inline.

java/pom.xml
robzor92: **Valid — should fix.** Confirmed the property split at head:

```
top-level     spark.version=4.1.3.0  scala.version=2.13.17  scala-short=2.13  hudi.groupId=io.hops.hudi  hudi.version=1.2.0
spark-4.1     all of the above (mirrors top level)
spark-3.5     spark.version=3.5.5.2, artifact.spark.version=spark3.5          <- inherits scala 2.13 + hudi 1.2.0
spark-3.3     artifact.spark.version=spark3.3                                 <- inherits everything else
```

So the structural claim holds: both older profiles now inherit a Scala and Hudi stack they were never meant to build against.

I could not confirm the specific "does not exist" assertions from outside — the Hops nexus paths I tried for `hudi-spark3.5-bundle` returned 404 for both `_2.12` and `_2.13`, so that probe does not discriminate. That does not change the conclusion: a profile that half-describes its stack is a trap regardless of which coordinate happens to resolve today. Deleting them if 5.1 no longer builds 3.3/3.5 is the cleaner of your two options.

python/hsfs/core/iceberg_engine.py
robzor92: **Valid — should fix (verify).** Confirmed the formatter at `hopsworks_common/util.py:278`:

```python
datetime.fromtimestamp(timestamp / 1000, timezone.utc).strftime("%Y-%m-%d %H:%M:%S.%f")[:-3] + "Z"
```

so the emitted string is `2026-08-28 10:15:30.123Z` — space separator, trailing `Z`, not ISO-8601, and shaped for Delta.

Agreed the failure mode is the dangerous part: an unrecognised read option is ignored rather than rejected, so a wrong format reads the latest snapshot and the test — which asserts dict equality against a mock — passes either way. Strongly agree these should not merge independently of loadtest#997, which is the only thing in the train that would actually exercise it. Worth noting loadtest#997 is green and approved now, so the ordering is available.

python/hsfs/core/iceberg_engine.py
robzor92: **Valid — should fix (docs).** Confirmed. The rename was applied mechanically to the prose, and the paragraph now argues against the very option the `if not snapshots:` branch below it passes. Before the rename it was a correct argument about `as-of-timestamp`.

Worth fixing together with the formatter question on line 1034, since whatever is decided there determines what this paragraph should say.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/FeatureGroup.java
robzor92: **Valid — should fix (small).** Confirmed the shim appears four times (`FeatureGroup`, `ExternalFeatureGroup`, `TrainingDataset`, and `StreamFeatureGroup` in hsfs/beam/flink), all empty, none explained.

The delombok output path also moved in this PR (`${project.basedir}/delombok` → `${project.build.directory}/generated-sources/delombok`), which supports the doclint theory. One line on one of them, referencing the other three, would settle it for the next reader.

java/pom.xml
robzor92: **Fixed** in `0a846504` — both profiles deleted.

Took deletion over completing them, for the reason you gave: 5.1 ships only Spark 4.1, so a profile
nothing builds and nothing tests would drift the same way at the next bump.

Worth recording the mechanism, because it is a good trap to know about: these profiles **worked
before this PR**. The top-level properties used to be the 3.5 stack, so what they left unset was
inherited correctly. The upgrade did not break them by editing them — it broke them by moving the
defaults out from under them. Any half-populated profile has that property.

XML parses, `mvn -o -N validate` passes, remaining profiles are `spark-4.1` and `with-hops-ee`, and
nothing else in the repo references either deleted id.

PR #1054, opened 2026-07-13
[FSTORE-2066] Fixes for stats computation / feature monitoring

javierdlrm:
## Summary
- Strip uniqueness-family metrics (uniqueness, distinctness, entropy, exactNumDistinctValues) from parsed deequ statistics when `exact_uniqueness` is disabled, so they deserialize as `None` instead of the JVM profiler's spurious `0.0`; guard the feature monitoring difference computation against `None` metric values.
- Resolve end-bounded Iceberg time-travel reads to a `snapshot-id` client-side, against the snapshots' own `committed_at` clock (the clock the backend derives the recorded commit time from), instead of passing `as-of-timestamp`. Passing the timestamp let Iceberg resolve the bound server-side against the snapshot-log clock, so a recorded commit time landing just below snapshot N's log timestamp could silently read snapshot N-1 (the mid-range skew), and a fresh-insert commit time landing before the first snapshot made `compute_statistics` die with "Cannot find a snapshot older than ...". Resolving here shares one clock, so both cases are handled. When the end bound precedes every snapshot the earliest snapshot is pinned (mirrors the Delta `versionAsOf` fallback); note this reads data committed after the window when the window genuinely ended before the first commit, rather than returning empty.
- Scrub non-finite values (NaN/Infinity) from client-side statistics serialization: `FeatureDescriptiveStatistics.to_dict()` maps them to `None` (all numeric fields + percentiles), and the pandas statistics conversion skips NaN mean/sum/min/max/percentiles (previously only `std` was guarded). Defense-in-depth alongside the backend sanitization in logicalclocks/hopsworks-ee (same ticket branch).

- Omit the uniqueness-family keys from the Java profiler output (`ColumnProfiler`/`ProfileJsonSerializer`) when exact uniqueness is disabled. Cluster verification showed the Python-side strip does not cover HUDI feature groups: their materialization job is the Scala/Java Spark job, which profiles and registers statistics through the Java client. All parsers copy only present keys, so absent keys deserialize as null on every path. The per-column entropy groupBy is also skipped when the flag is off.
- Compute uniqueness from the exact singleton count in the Java profiler: the shortcut `(2*distinct - n)/n` equals Deequ's definition only when no value occurs more than twice; the singleton count now comes from the same frequency pass as entropy.
- Resolve end-bounded Delta reads to a version via the Delta log's in-commit `commitInfo.timestamp` values and pin `versionAsOf`. Two layers fixed on-cluster: `timestampAsOf` resolves against commit-file modification times, and `DeltaTable.history()` also carries modtimes — both lag the in-commit timestamps the backend records as commit times, so end-bounded reads could silently land on the previous version (statistics for commit N computed over commit N-1's data).
- Read Delta commit timestamps from `_delta_log/*.json` (`commitInfo.timestamp`) instead of `DeltaTable.history()`: history() carries commit-file modification times, which lag the in-commit timestamps the backend records, so the version resolution could still land on the previous version.
- Note: the FSTORE-2067 vectorization workarounds (table-creation property and statistics/monitoring read option) live entirely in #1067 now. That PR must merge first: without it, Iceberg statistics and monitoring reads and training dataset creation hit the vectorized-reader executor crash, so the ICEBERG variants of this PR's fixes only work on top of #1067.
- Fall back to an end-bounded snapshot read when an incremental Iceberg window starts before the first snapshot (rolling monitoring windows on fresh tables previously died on the incremental-branch raise).

## Test plan
- [x] Unit: `_parse_deequ_statistics` strips uniqueness metrics with the flag off, preserves them with it on
- [x] Unit: monitoring diff returns `None` when a metric is missing on either side
- [x] Unit: Iceberg end-only read opts — resolvable end bound pins the snapshot at/before it (mid-range case), pre-first-snapshot falls back to earliest `snapshot-id`, no snapshots keeps timestamp bound
- [x] Unit: `to_dict()` NaN/Infinity scrub (fields, percentile list and mapping); pandas conversion of all-null column emits only dataType+count
- [x] Full suite: 3649 passed (5 pre-existing avro/spark environment failures, confirmed failing on clean tree)
- [ ] Integration (pending cluster): `test_statistics[DELTA/HUDI]`, `test_batch[ICEBERG]`, `test_feature_monitoring[ICEBERG]`, `fs_statistics_wide_tables` variants
- [x] Unit (java/spark): profiler omits uniqueness/distinctness/entropy/exactNumDistinctValues when exactUniqueness=false, emits them when true; parity tests unchanged
- [x] Unit (java/spark): uniqueness equals the exact singleton fraction with multiplicity >= 3; profiler parity tests unchanged
- [x] Unit: Delta end-bounded read opts resolve the version between commits, at the earliest commit, and fall back to the timestamp bound when the log is unreadable
- [x] Integration: `test_statistics.py` `test_workflow[DELTA]` + `[HUDI]` pass end-to-end on a cluster (2 passed)
- [x] Integration: `test_feature_monitoring.py` `[DELTA]` + `[HUDI]` + `[ICEBERG]` pass end-to-end on a cluster (3 passed)

https://hopsworks.atlassian.net/browse/FSTORE-2066

### DIFF
--- a/java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/profile/ColumnProfiler.java
+++ b/java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/profile/ColumnProfiler.java
@@ -63,9 +63,11 @@ import java.util.Map;
  *
  * <h3>Uniqueness formula</h3>
  *
- * <p>{@code uniqueness = max(0, (2 * exactDistinct - nonNull) / nonNull)} — verified against
- * the Deequ baseline. This formula holds because Deequ defines uniqueness as the fraction of
- * values appearing exactly once, and with random integers most duplicates appear exactly twice.
+ * <p>{@code uniqueness = singletons / nonNull} — Deequ's exact definition (fraction of values
+ * appearing exactly once). The singleton count comes from the same per-value frequency pass
+ * as entropy, so no additional Spark job is paid for it. (An earlier shortcut,
+ * {@code (2 * exactDistinct - nonNull) / nonNull}, is only equivalent when no value occurs
+ * more than twice and undercounts otherwise.)
  *
  * <h3>stdDev</h3>
  *
@@ -116,10 +118,12 @@ public class ColumnProfiler {
       }
 
       Row aggRow = computeScalars(df, columns, exactUniqueness);
-      // entropy is derived from exact per-value frequencies; only pay for the
-      // per-column groupBy when exact uniqueness stats were requested
-      Map<String, Double> entropyMap = exactUniqueness
-          ? computeEntropy(df, columns) : new LinkedHashMap<String, Double>();
+      // entropy and the uniqueness singleton count are derived from exact per-value
+      // frequencies; only pay for the per-column groupBy when exact uniqueness
+      // stats were requested
+      Map<String, ValueFrequencyStats> frequencyStatsMap = exactUniqueness
+          ? computeValueFrequencyStats(df, columns)
+          : new LinkedHashMap<String, ValueFrequencyStats>();
 
       Map<String, Map<String, Double>> correlationMap =
           new LinkedHashMap<String, Map<String, Double>>();
@@ -129,7 +133,7 @@ public class ColumnProfiler {
 
       List<ColumnProfile> profiles = new ArrayList<ColumnProfile>(columns.size());
       for (ColumnInfo col : columns) {
-        ColumnProfile cp = buildProfile(df, col, aggRow, entropyMap, correlationMap,
+        ColumnProfile cp = buildProfile(df, col, aggRow, frequencyStatsMap, correlationMap,
             histogram, histogramBins, kll, exactUniqueness);
         profiles.add(cp);
       }
@@ -213,18 +217,31 @@ public class ColumnProfiler {
   }
 
   // ---------------------------------------------------------------------------
-  // Pass 2: entropy — one groupBy per column
+  // Pass 2: per-value frequencies (entropy + uniqueness singletons) — one groupBy per column
   // ---------------------------------------------------------------------------
 
-  private Map<String, Double> computeEntropy(Dataset<Row> df, List<ColumnInfo> columns) {
-    Map<String, Double> result = new LinkedHashMap<String, Double>();
+  /** Per-column stats derived from the exact per-value frequency distribution. */
+  private static final class ValueFrequencyStats {
+    private final double entropy;
+    private final long singletons;
+
+    private ValueFrequencyStats(double entropy, long singletons) {
+      this.entropy = entropy;
+      this.singletons = singletons;
+    }
+  }
+
+  private Map<String, ValueFrequencyStats> computeValueFrequencyStats(Dataset<Row> df,
+      List<ColumnInfo> columns) {
+    Map<String, ValueFrequencyStats> result = new LinkedHashMap<String, ValueFrequencyStats>();
     for (ColumnInfo col : columns) {
-      result.put(col.name, computeColumnEntropy(df, col.name));
+      result.put(col.name, computeColumnValueFrequencyStats(df, col.name));
     }
     return result;
   }
 
-  private double computeColumnEntropy(Dataset<Row> df, String columnName) {
+  private ValueFrequencyStats computeColumnValueFrequencyStats(Dataset<Row> df,
+      String columnName) {
     Column cc = functions.col(columnName);
     List<Row> rows = df.filter(cc.isNotNull())
         .groupBy(cc)
@@ -233,14 +250,19 @@ public class ColumnProfiler {
         .collectAsList();
 
     if (rows.isEmpty()) {
-      return 0.0;
+      return new ValueFrequencyStats(0.0, 0L);
     }
     long total = 0;
+    long singletons = 0;
     for (Row row : rows) {
-      total += row.getLong(0);
+      long count = row.getLong(0);
+      total += count;
+      if (count == 1) {
+        singletons++;
+      }
     }
     if (total == 0) {
-      return 0.0;
+      return new ValueFrequencyStats(0.0, 0L);
     }
     double entropy = 0.0;
     for (Row row : rows) {
@@ -250,7 +272,7 @@ public class ColumnProfiler {
         entropy -= pp * Math.log(pp);
       }
     }
-    return entropy;
+    return new ValueFrequencyStats(entropy, singletons);
   }
 
   // ---------------------------------------------------------------------------
@@ -285,7 +307,7 @@ public class ColumnProfiler {
   // ---------------------------------------------------------------------------
 
   private ColumnProfile buildProfile(Dataset<Row> df, ColumnInfo col, Row aggRow,
-      Map<String, Double> entropyMap,
+      Map<String, ValueFrequencyStats> frequencyStatsMap,
       Map<String, Map<String, Double>> correlationMap,
       boolean histogram, int histogramBins, boolean kll, boolean exactUniqueness) {
     String nn = col.name;
@@ -300,15 +322,16 @@ public class ColumnProfiler {
     // meaningful when exactUniqueness=true. When false they stay null and the serializer
     // omits their keys, so consumers deserialize them as absent instead of a bogus 0.
     Long exactDistinct = exactUniqueness ? (Long) aggRow.getAs(nn + "__exact_distinct") : null;
+    ValueFrequencyStats freqStats = frequencyStatsMap.get(nn);
 
     double completeness = total > 0 ? (double) nonNull / total : 0.0;
     Double distinctness = exactUniqueness
         ? (nonNull > 0 ? (double) exactDistinct / nonNull : 0.0) : null;
     Double uniqueness = exactUniqueness
-        ? (nonNull > 0 ? Math.max(0.0, (double) (2L * exactDistinct - nonNull) / nonNull) : 0.0)
+        ? (nonNull > 0 && freqStats != null ? (double) freqStats.singletons / nonNull : 0.0)
         : null;
     Double entropy = exactUniqueness
-        ? (entropyMap.containsKey(nn) ? entropyMap.get(nn) : 0.0) : null;
+        ? (freqStats != null ? freqStats.entropy : 0.0) : null;
 
     ColumnProfile.Builder builder = new ColumnProfile.Builder()
         .columnName(nn)
--- a/java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/profile/ColumnProfiler.java
+++ b/java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/profile/ColumnProfiler.java
@@ -116,7 +116,10 @@ public class ColumnProfiler {
       }
 
       Row aggRow = computeScalars(df, columns, exactUniqueness);
-      Map<String, Double> entropyMap = computeEntropy(df, columns);
+      // entropy is derived from exact per-value frequencies; only pay for the
+      // per-column groupBy when exact uniqueness stats were requested
+      Map<String, Double> entropyMap = exactUniqueness
+          ? computeEntropy(df, columns) : new LinkedHashMap<String, Double>();
 
       Map<String, Map<String, Double>> correlationMap =
           new LinkedHashMap<String, Map<String, Double>>();
@@ -294,18 +297,18 @@ public class ColumnProfiler {
     long total = nonNull + nullCount;
     long approxDistinct = aggRow.getAs(nn + "__approx_distinct");
     // exactNumDistinctValues + derived stats (distinctness/uniqueness/entropy) are only
-    // meaningful when exactUniqueness=true. When false we leave them at 0; the serializer
-    // still emits the keys (matching Deequ's shape, which is always keys-present).
-    long exactDistinct = exactUniqueness ? (Long) aggRow.getAs(nn + "__exact_distinct") : 0L;
+    // meaningful when exactUniqueness=true. When false they stay null and the serializer
+    // omits their keys, so consumers deserialize them as absent instead of a bogus 0.
+    Long exactDistinct = exactUniqueness ? (Long) aggRow.getAs(nn + "__exact_distinct") : null;
 
     double completeness = total > 0 ? (double) nonNull / total : 0.0;
-    double distinctness = exactUniqueness && nonNull > 0
-        ? (double) exactDistinct / nonNull : 0.0;
-    double uniqueness = exactUniqueness && nonNull > 0
-        ? Math.max(0.0, (double) (2L * exactDistinct - nonNull) / nonNull)
-        : 0.0;
-    double entropy = exactUniqueness && entropyMap.containsKey(nn)
-        ? entropyMap.get(nn) : 0.0;
+    Double distinctness = exactUniqueness
+        ? (nonNull > 0 ? (double) exactDistinct / nonNull : 0.0) : null;
+    Double uniqueness = exactUniqueness
+        ? (nonNull > 0 ? Math.max(0.0, (double) (2L * exactDistinct - nonNull) / nonNull) : 0.0)
+        : null;
+    Double entropy = exactUniqueness
+        ? (entropyMap.containsKey(nn) ? entropyMap.get(nn) : 0.0) : null;
 
     ColumnProfile.Builder builder = new ColumnProfile.Builder()
         .columnName(nn)
--- a/java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/profile/ColumnProfiler.java
+++ b/java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/profile/ColumnProfiler.java
@@ -50,29 +50,33 @@ import java.util.Map;
  * <h3>KLL gating divergence from Deequ</h3>
  *
  * <p>Deequ always emits {@code kll} for numeric columns regardless of {@code withKLLProfiling()};
- * the toggle is a no-op in 2.0.7-spark-3.5. This implementation emits {@code kll} only when
- * {@code kll=true}, aligning with the Phase-1 API contract. The {@code kll=false} path produces
- * smaller profiles. The golden-parity test (task #6) accounts for this known divergence.
+ * the toggle is a no-op in 2.0.7-spark-3.5.
+ * This implementation emits {@code kll} only when {@code kll=true}, aligning with the Phase-1
+ * API contract.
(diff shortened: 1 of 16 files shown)


python/hsfs/core/statistics_engine.py
```diff
@@ -600,7 +605,9 @@ def _save_statistics(
             )
         return stats
 
-    def _parse_deequ_statistics(self, stats) -> list[FeatureDescriptiveStatistics]:
+    def _parse_deequ_statistics(
+        self, stats, exact_uniqueness: bool
+    ) -> list[FeatureDescriptiveStatistics]:
         if stats is None:
             warnings.warn(
```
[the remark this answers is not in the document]
javierdlrm: Fixed in aef05994 — annotated as `list[FeatureDescriptiveStatistics] | None`.

python/hsfs/core/feature_descriptive_statistics.py
```diff
@@ -27,6 +28,17 @@
     from collections.abc import Mapping
 
 
+def _finite_or_none(value):
+    """Map non-finite numbers (NaN, Infinity) to None.
+
+    Non-finite values are not valid JSON and not valid SQL numerics: a single
+    one aborts the whole statistics registration on the backend.
```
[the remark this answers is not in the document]
javierdlrm: Fixed in aef05994 — the scrub now accepts any `numbers.Real` (numpy float32 scalars are not `float` instances; numpy registers its scalar types with the numbers ABC).

python/hsfs/core/iceberg_engine.py
```diff
@@ -834,11 +835,29 @@ def _setup_iceberg_read_opts(
...
+                    iceberg_options = {
+                        self.ICEBERG_QUERY_TIME_TRAVEL_SNAPSHOT_ID: str(
+                            snapshots[0]["snapshot_id"]
+                        ),
+                    }
+                else:
+                    # no snapshots at all; keep the timestamp bound
+                    iceberg_options = {
+                        self.ICEBERG_QUERY_TIME_TRAVEL_AS_OF_TIMESTAMP: str(end_ts),
+                    }
```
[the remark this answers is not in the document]
javierdlrm: Fixed in aef05994 — the snapshot log is read once and the earliest-snapshot commit time decides directly between as-of-timestamp and snapshot-id, mirroring the Delta earliest-commit comparison.

python/hsfs/core/iceberg_engine.py
```diff
@@ -876,6 +897,26 @@ def _setup_iceberg_read_opts(
...
 
+    def _snapshot_read_opts_at(self, location: str, end_ts: int) -> dict[str, str]:
+        """Read options pinning the table state at *end_ts*.
+
+        A time that predates the table's first snapshot pins the earliest
+        snapshot by id instead: compute_statistics runs right after a fresh
+        insert, and the backend-recorded commit_time can be a few ms before
+        the first snapshot's commit, a range where Iceberg rejects
+        as-of-timestamp.
+        """
```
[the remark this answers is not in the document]
javierdlrm: Fixed in a65de0b3b — one sentence per line.

python/hsfs/core/monitoring_window_config_engine.py
```diff
@@ -558,7 +558,7 @@ def _fetch_feature_view_data(
         else:
             entity_df = fv_query.as_of(
                 exclude_until=start_time, wallclock_time=end_time
-            ).read()
+            ).read(read_options=statistics_engine._ICEBERG_STATS_READ_OPTIONS)
```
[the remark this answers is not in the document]
javierdlrm: Keeping as-is: cross-module use of underscore-prefixed members between internal hsfs.core modules is the established idiom in this codebase (e.g. util._convert_event_time_to_timestamp, statistics_engine internals), and the constant is deliberately private because it is a temporary FSTORE-2067 workaround slated for removal.

python/hsfs/core/delta_engine.py
```diff
@@ -316,43 +325,72 @@ def _setup_delta_read_opts(
...
+                from pyspark.sql import functions as F  # noqa: PLC0415
 
-                rows = (
-                    DeltaTable.forPath(self._spark_session, location)
-                    .history()
-                    .select("version", "timestamp")
-                    .collect()
+                log_df = self._spark_session.read.json(
+                    location.rstrip("/") + "/_delta_log/*.json"
                 )
```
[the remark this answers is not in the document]
javierdlrm: Fixed in 16ec702a6 — the log read now uses an explicit schema projecting only commitInfo.timestamp and filters null timestamps, avoiding the add/remove payload parsing and the TypeError-to-fallback degradation.

python/hsfs/core/delta_engine.py
```diff
@@ -271,17 +271,26 @@ def _setup_delta_read_opts(
             delta_fg_alias.left_feature_group_end_timestamp is not None
             and delta_fg_alias.left_feature_group_start_timestamp is None
         ):
-            # snapshot query with end time
+            # snapshot query with end time; resolve the version against the Delta
+            # log's in-commit timestamps instead of passing timestampAsOf. Delta
+            # resolves timestamp bounds against commit-file modification times,
```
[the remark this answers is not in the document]
javierdlrm: Fixed in 6ac64a79b: reflowed to one sentence per line.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/profile/ProfileJsonSerializer.java
```diff
@@ -123,17 +122,34 @@ private Map<String, Object> toCategoricalMap(ColumnProfile profile) {
...
       map.put("histogram", profile.getHistogram());
     }
     return map;
   }
 
+  /**
+   * Emits the uniqueness-family stats, omitting the keys when they were not computed
+   * (exactUniqueness disabled) so consumers deserialize them as absent instead of a
+   * bogus 0. Divergence from Deequ's always-keys-present shape is deliberate.
+   */
```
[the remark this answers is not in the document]
javierdlrm: Fixed in 6ac64a79b: reflowed to one sentence per line.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/profile/ColumnProfiler.java
```diff
@@ -63,9 +63,11 @@
...
  * <h3>Uniqueness formula</h3>
  *
- * <p>{@code uniqueness = max(0, (2 * exactDistinct - nonNull) / nonNull)} — verified against
- * the Deequ baseline. This formula holds because Deequ defines uniqueness as the fraction of
- * values appearing exactly once, and with random integers most duplicates appear exactly twice.
+ * <p>{@code uniqueness = singletons / nonNull} — Deequ's exact definition (fraction of values
+ * appearing exactly once). The singleton count comes from the same per-value frequency pass
+ * as entropy, so no additional Spark job is paid for it. (An earlier shortcut,
+ * {@code (2 * exactDistinct - nonNull) / nonNull}, is only equivalent when no value occurs
+ * more than twice and undercounts otherwise.)
```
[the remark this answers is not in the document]
javierdlrm: Fixed in 6ac64a79b: reflowed the whole class Javadoc to one sentence per line.

javierdlrm (commented):
Tests please

python/hsfs/core/iceberg_engine.py
```diff
@@ -995,6 +1006,24 @@ def _setup_iceberg_read_opts(
...
+        snapshots = self._read_snapshots(location)
+        if snapshots and (
+            util._convert_event_time_to_timestamp(snapshots[0]["committed_at"]) > end_ts
+        ):
+            return {
+                self.ICEBERG_QUERY_TIME_TRAVEL_SNAPSHOT_ID: str(
+                    snapshots[0]["snapshot_id"]
+                ),
+            }
+        return {self.ICEBERG_QUERY_TIME_TRAVEL_AS_OF_TIMESTAMP: str(end_ts)}
```
robzor92: This fallback only covers the case where the skew is *observable* (`end_ts` before the first snapshot → Iceberg throws). The same skew at commit N>1 is silent: `as-of-timestamp` resolves against `table.history()` entries with an inclusive `<=` (`SnapshotUtil.snapshotIdAsOfTime`), so an `end_ts` that lands just below snapshot N's log timestamp reads snapshot **N−1** — statistics for commit N computed over commit N−1's data, i.e. exactly the silent case this PR fixes for Delta by resolving versions client-side.

The fresh-insert failure you observed proves the registered commit time can land below the resolution clock, so the mid-range variant is possible too (intermittent, which is why `test_feature_monitoring[ICEBERG]` can still pass).

Suggestion: mirror the Delta fix — resolve client-side with the same conversion used at registration (latest snapshot with `_convert_event_time_to_timestamp(committed_at) <= end_ts`, else earliest; `_resolve_snapshot_id_at` already implements this) and always pin `snapshot-id` instead of passing `as-of-timestamp`. Registration and resolution then share a clock, and mid-range resolution becomes exact.

Related: I couldn't reproduce the "ms-floored" mechanism from the description — `int(dt.timestamp() * 1000)` is exact for whole-ms datetimes (brute-forced 500k values, 0 mismatches), and the backend stores the client's ms `commitTime` verbatim. The skew more likely lives between snapshot `committed_at` and the snapshot-log entry timestamps `history()` returns. Worth pinning down, since the fix depends on which clocks agree.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/profile/ProfileJsonSerializer.java
```diff
@@ -89,11 +92,7 @@ private Map<String, Object> toNumericMap(ColumnProfile profile) {
     map.put("completeness", profile.getCompleteness());
     map.put("numRecordsNonNull", profile.getNumRecordsNonNull());
     map.put("numRecordsNull", profile.getNumRecordsNull());
-    map.put("distinctness", profile.getDistinctness());
-    map.put("entropy", profile.getEntropy());
-    map.put("uniqueness", profile.getUniqueness());
-    map.put("approximateNumDistinctValues", profile.getApproximateNumDistinctValues());
-    map.put("exactNumDistinctValues", profile.getExactNumDistinctValues());
+    putUniquenessFamily(map, profile);
     map.put("mean", profile.getMean());
```
robzor92: An all-null numeric column still kills registration on this path: Spark's `avg`/`min`/`max`/`sum`/`stddev_pop` over an all-null column return null, these puts emit `"mean": null` etc. (Jackson writes map nulls by default), and `FeatureDescriptiveStatistics.fromDeequStatisticsJson` then does `has("minimum") → getDouble(...)` — org.json's `has()` returns true for an explicit JSON null and `getDouble` throws on `JSONObject.NULL`. So the Java client crashes parsing its own profile output.

The wide-table fixture has `long_column_all_null`, so I'd expect `fs_statistics_wide_tables[HUDI]` (the unchecked integration box) to still fail.

Suggestion: omit null-valued keys here, same as `putUniquenessFamily` does — "absent = not computed" is now the established contract anyway. (Or `optDouble`/`isNull` guards in the parser; the backend's `buildFromDeequJson` has the same `has→getDouble` hazard on the content path.)

python/hsfs/core/delta_engine.py
```diff
@@ -317,43 +326,85 @@ def _setup_delta_read_opts(
...
+        """Get all commits from the Delta log.
 
-        Return (version, timestamp_ms) of the earliest commit in the Delta
-        log at `location`, or None if the history cannot be read.
+        Return a list of (version, timestamp_ms) tuples sorted by version
+        ascending, or None if the log cannot be read or is empty.
+
+        The timestamps are the log's in-commit ``commitInfo.timestamp`` values,
+        the same clock the backend-recorded commit times are derived from.
+        ``DeltaTable.history()`` is deliberately NOT used here: its timestamp
```
robzor92, line 348: This docstring is correct — I checked delta-spark 3.3 `DeltaHistoryManager`: without in-commit timestamps enabled, both `history()` and `timestampAsOf` use commit-file modification times. But it directly contradicts the pre-existing comment at the top of the CDF retry above ("...while DeltaTable.history() returns the in-commit creation timestamp"). Please fix that stale comment while here.

Corollary worth stating in this docstring: commits registered via classic Spark carry `history()` = modtime, not the in-commit clock (only delta-rs and the Spark Connect path register in-commit values). The resolution below is still correct for those — `ict(N) <= modtime(N) < ict(N+1)` for sequential writers, so `commit_ts <= end_ts` clamps to the intended version — but that's the argument that makes the pyspark-driver variants safe, and those are exactly the ones not yet cluster-validated.

robzor92 (commented):
Did a cross-repo pass over this + logicalclocks/hopsworks-ee#3169 + logicalclocks/loadtest#958, tracing the contracts end-to-end (profiler → parsers → backend → loadtest expectations). The Delta version resolution, the uniqueness singleton formula (loadtest's expected `0.266667` = 4/15 confirms both engines now agree on Deequ's definition), and the NaN scrub layering all hold up under adversarial reading.

Two things I'd address before merge (details inline):

1. **Iceberg end-bounded fallback only fixes the before-first-snapshot case.** The same skew mid-range silently resolves to snapshot N−1 — the exact silent bug class the Delta half of this PR fixes. Client-side snapshot-id pinning closes it.
2. **All-null numeric columns still crash the Java client's own profile parse** (`"mean": null` → org.json `getDouble` throws). The wide-table fixture has `long_column_all_null`, so I'd expect `fs_statistics_wide_tables[HUDI]` (the unchecked integration box) to still fail as-is.

Also noting: no CI checks have run on this branch yet, and per the description #1067 must merge first for the ICEBERG variants — worth sequencing explicitly.

One trade-off to be aware of (fine to keep, maybe worth a comment): both the Delta earliest-version pin and the new Iceberg earliest-snapshot pin read data committed *after* the window when `end_ts` genuinely predates the first commit (e.g. a monitoring window that ended before the table's first insert). That's indistinguishable client-side from the skew case, but it means detection windows can silently include future data instead of coming back empty.

python/hsfs/core/delta_engine.py
```diff
@@ -271,17 +271,26 @@ def _setup_delta_read_opts(
...
-                # insert: the backend-recorded commit_time can be a few ms before
-                # the Delta log's first commit, and Delta rejects timestampAsOf
-                # in that range. Fall back to versionAsOf on the earliest commit.
+            commits = self._get_delta_commits(location) if location else None
+            if commits:
+                # latest version committed at or before end_ts; when end_ts
+                # predates the first history timestamp, pin the earliest version
+                version = commits[0][0]
+                for commit_version, commit_ts in commits:
+                    if commit_ts <= end_ts:
```
robzor92: commitInfo timestamps aren't guaranteed monotonic in version (concurrent writers / commit retries — Delta itself monotonizes when resolving). With out-of-order timestamps this loop can pick a higher version even though an intermediate commit is newer than `end_ts`. Cheap hardening: break on the first `commit_ts > end_ts`.

python/hsfs/core/delta_engine.py
```diff
@@ -317,43 +326,85 @@ def _setup_delta_read_opts(
...
+                            StructType([StructField("timestamp", LongType())]),
+                        )
+                    ]
+                )
                 rows = (
-                    DeltaTable.forPath(self._spark_session, location)
-                    .history()
-                    .select("version", "timestamp")
+                    self._spark_session.read.schema(log_schema)
+                    .json(location.rstrip("/") + "/_delta_log/*.json")
```
robzor92, line 380: Minor: this glob also matches V2 checkpoint manifests (`<v>.checkpoint.<uuid>.json` — and `(\d+)\.json` would extract the uuid's trailing digits as the "version") and log-compaction files (`<x>.<y>.compacted.json`). Today the `commitInfo.timestamp` filter (or the broad except → timestamp fallback) saves it, but anchoring on the 20-digit basename (`(\d{20})\.json$`) would make it deterministic. Same idiom exists in the pre-existing Spark Connect path of `_get_last_commit_metadata`.

python/hsfs/core/iceberg_engine.py
javierdlrm: Fixed in 32f0e581a. Switched `_snapshot_read_opts_at` to resolve the snapshot id client-side (`_latest_snapshot_id_at`, the same committed_at clock as `_resolve_snapshot_id_at`) and always pin `snapshot-id`; `as-of-timestamp` is only used when the table has no snapshots at all. Registration and read now share one clock, so the mid-range N-1 case is closed, not just the before-first-snapshot case. You're also right about the mechanism: I dropped the ms-floor claim from the description; the skew is between the snapshot committed_at and the log timestamps history/as-of resolve against. The earliest-snapshot fallback and its future-data trade-off (window ending before the first commit reads post-window data) are now documented on the method, mirroring Delta.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/profile/ProfileJsonSerializer.java
javierdlrm: Fixed in 32f0e581a. `ProfileJsonSerializer` now omits null-valued mean/maximum/minimum/sum/stdDev (new `putIfNotNull`), same 'absent = not computed' contract as `putUniquenessFamily`, so an all-null column no longer emits `"mean": null`. That removes the crash at the source for the round-trip; I also guarded the backend content-path parser (`buildFromDeequJson`) with isNull in #3169 since that JSON can come from older clients.

python/hsfs/core/delta_engine.py
javierdlrm: Fixed in 32f0e581a: corrected the stale CDF-retry comment (history() reports the commit-file modification time, not the in-commit timestamp) and added the corollary to the `_get_delta_commits` docstring — reading commitInfo.timestamp from the log directly is what keeps resolution correct for classic-Spark-written tables where history() would be modtime.

python/hsfs/core/delta_engine.py
javierdlrm: Fixed in 32f0e581a: the loop now breaks on the first commit past end_ts instead of scanning all commits, so an out-of-order later commit can't select too high a version.

python/hsfs/core/delta_engine.py
javierdlrm: Fixed in 32f0e581a: version extraction now anchors on the 20-digit commit basename (`(\d{20})\.json$`) and filters out rows whose version didn't match, so V2 checkpoint manifests and log-compaction files are excluded deterministically rather than relying on the commitInfo.timestamp filter.

robzor92 (approved):

PR #1069, opened 2026-07-23
[FSTORE-2068] Delete online rows on remove_rows (deprecate commit_delete_record)

dhananjay-mk:
## What

Adds online-store deletion to feature-group record deletes and renames the
public method.

- `remove_rows(delete_df, write_options=None, delete_online=None)` is the new
  public method; `commit_delete_record` is kept as a deprecated alias that
  delegates to it, so existing callers keep working (with a deprecation warning).
  The alias stays offline-only by default, preserving the behaviour it had
  before the online delete existed.
- The online delete removes the rows from the online store: the SDK produces an
  `operation: delete` tombstone to the feature group's Kafka topic, which
  OnlineFS applies to RonDB by primary key.
- `delete_online` is tri-state. Left unset it follows the feature group, so an
  online-enabled feature group is deleted from both stores and one without an
  online store is deleted from offline only. `False` forces offline-only.
  `True` on a feature group that has no online store this release can delete
  from raises rather than silently skipping the leg.
- `delete_df` needs to carry only the primary key columns (null-padded to the
  feature group schema before encoding); any other columns are ignored for the
  online delete. The offline leg still needs whatever columns its merge matches
  on, which for a feature group with an `event_time` or partition columns is
  more than the primary key.
- Stream feature groups are supported for all of DELTA, ICEBERG and HUDI. Both
  materialization paths drop the tombstone so it is never replayed into the
  offline table: the Java `DeltaStreamerAvroDeserializer` returns null for a
  delete header (HUDI), and the PySpark `hsfs_utils` materialization filters it
  out (DELTA and ICEBERG, which the backend routes to the util).
- Feature groups with an embedding index are not supported, since their online
  data lives in the vector database.
- No backend-version gate: the online-delete path does not involve the backend,
  and OnlineFS capability is pinned to the cluster by helm. This SDK needs to be
  paired with a delete-capable OnlineFS.

Java SDK (FeatureGroup, StreamFeatureGroup) mirrors the rename and the
`deleteOnline` overloads.

Also carries the docstring updates for the `DELETED` online-ingestion status
added under FSTORE-2080, which is what the online delete is now reported as.

## Companion PRs

- clusterj-onlinefs #177: OnlineFS applies the delete tombstone to RonDB.
- loadtest: online-delete workflow tests.

## Testing

Python + Java unit tests (delegation, the unset/explicit resolution of
`delete_online`, PK-only serialization, tombstone filtering in both
materialization paths).

Validated end-to-end on a cluster via the loadtest online-delete tests. The
non-stream case removes the row from both stores. The stream cases were run on
both engines with the materialization job forced between the insert and the
delete, and the row stays deleted after the post-delete materialization run for
DELTA and HUDI. ICEBERG goes through the same PySpark materialization path as
DELTA and is covered by unit tests, but has not had its own cluster run.

Note on the stream test shape: the materialization job has to run between the
insert and the delete. On the Spark engine `insert(wait_for_job=True)` does not
materialize, so without an explicit run the post-delete materialization replays
the inserts and both a fixed and an unfixed build appear to resurrect the row.


python/hsfs/engine/spark.py
```diff
@@ -832,17 +832,112 @@ def _save_online_dataframe(self, feature_group, dataframe, write_options):
...
+        dataframe = self._pad_online_delete_dataframe(feature_group, dataframe)
+        serialized_df = self._serialize_to_avro(feature_group, dataframe)
+
+        (
+            serialized_df.withColumn(
+                "headers",
+                self._get_headers(
+                    feature_group, dataframe.count(), write_options, operation="delete"
+                ),
+            )
```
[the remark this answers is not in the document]
dhananjay-mk: Not applying this one, and I think the premise is inverted for the delete path.

`num_entries` has exactly one consumer: `kafka_engine._get_headers` passes it to `_create_online_ingestion` as the expected row total. On the insert path that total is optional metadata, which is why `disable_online_ingestion_count` exists (FSTORE-1978). On the delete path it is the completion signal: `OnlineIngestion.wait_for_completion` breaks only on

```python
if self.num_entries and rows_processed >= self.num_entries:
```

and `_delete_online_dataframe` calls `wait_for_completion` whenever `wait_for_online_ingestion` is set. Honoring the flag here would send `num_entries=None`, leaving that `if` unable to fire, so every waited online delete would block for the full timeout (60s default) and end with a spurious "Timeout ... was exceeded" warning instead of returning as soon as the tombstones land.

The tombstones do get counted on the consumer side, so this is not insert-only bookkeeping. In clusterj-onlinefs, `RondbCommitter.persistRow` handles the delete and returns `true`, which increments `committedRows`, and `CommitterInterface` reports that unconditionally via `OnlineIngestionManager.save(..., Status.UPSERTED, commitedRows)`. `rows_processed` therefore climbs for deletes, and `num_entries` is what it is compared against.

Agreed that `dataframe.count()` is a real Spark action, but a `remove_rows` frame is a list of primary keys, so the cost is not comparable to counting an insert dataset. Documented on `remove_rows` instead that `disable_online_ingestion_count` has no effect there, and why.

python/hsfs/engine/python.py
```diff
@@ -1971,6 +1971,73 @@ def _write_dataframe_kafka(
...
+        # primary key and ignores the values. `storage` b"1" routes it to the
+        # online store. `dataframe` needs to carry only the primary key: non-key
+        # fields are filled with null so the record serializes.
+        fill_values = kafka_engine._online_delete_fill_values(feature_group)
+        n_rows = len(dataframe)
+        producer, headers, feature_writers, writer = kafka_engine._get_kafka_resources(
+            feature_group,
+            offline_write_options,
+            num_entries=n_rows,
+        )
```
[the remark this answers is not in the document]
dhananjay-mk: Same reasoning as the `spark.py` thread, plus two corrections specific to this one.

`n_rows` here is `len(dataframe)` on a pandas or polars frame, which is O(1). There is no count to avoid on this path, so the cost argument does not apply.

`num_entries` also does not reach the Kafka producer. `_init_kafka_producer` takes only `feature_store_id` and `options`; the value flows solely into `_get_headers` and from there into the `OnlineIngestion` record. (The "Kafka producer configuration" phrasing appears in an existing comment in `_write_dataframe_kafka`, which is inaccurate for the same reason.)

What `n_rows` does drive is the online-ingestion record's expected total, which is the sole break condition in `OnlineIngestion.wait_for_completion`, and the local tqdm upload bar. `_delete_dataframe_kafka` calls `wait_for_completion` when `wait_for_online_ingestion` is set, so suppressing the count would turn every waited online delete into a full-timeout block plus a spurious timeout warning. OnlineFS does report tombstones against that record: the delete branch in `RondbCommitter.persistRow` returns `true`, incrementing `committedRows`, which `CommitterInterface` reports as `Status.UPSERTED`.

Leaving both delete paths counting, and documenting on `remove_rows` that `disable_online_ingestion_count` has no effect there because the wait depends on the count.

python/hsfs/engine/spark.py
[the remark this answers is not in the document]
dhananjay-mk: Correcting my previous reply: applied in 883702dd5.

My argument rested on a delete frame being a handful of primary keys, where the count is nearly free. That does not hold for a large `delete_df` on the Spark engine, which is the case `disable_online_ingestion_count` was added for on the insert path (FSTORE-1978). `dataframe.count()` is one action and `.save()` is another over the same lineage, with nothing cached in between, so an uncached source plan is evaluated twice. A caller who never sets `wait_for_online_ingestion` pays that for a total nothing reads back.

The completion-detection point stands, but it only applies when the caller opts into waiting, which is their explicit choice and already the documented trade on insert. It should not force the count on someone who never asked to wait.

All three delete paths now gate the count the same way the insert paths do: `python.py:_delete_dataframe_kafka`, `spark.py:_delete_online_dataframe`, and `SparkEngine.deleteOnlineDataframe`. `remove_rows` gained the `write_options` documentation it lacked, including that disabling the count means `wait_for_online_ingestion` runs to `timeout`. Unit tests cover the gate on the Python and Spark engines.

One thing this surfaced, pre-existing and not limited to deletes: disabling the count while setting `timeout` to `0` leaves neither break in `wait_for_completion` reachable (`online_ingestion.py:207,211`, and `OnlineIngestion.java:83,88`), so the wait hangs rather than timing out. Noted on every docstring carrying the option; worth a proper fix separately.

python/hsfs/engine/python.py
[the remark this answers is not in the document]
dhananjay-mk: Correcting my previous reply: applied in 883702dd5.

My argument rested on a delete frame being a handful of primary keys, where the count is nearly free. That does not hold for a large `delete_df` on the Spark engine, which is the case `disable_online_ingestion_count` was added for on the insert path (FSTORE-1978). `dataframe.count()` is one action and `.save()` is another over the same lineage, with nothing cached in between, so an uncached source plan is evaluated twice. A caller who never sets `wait_for_online_ingestion` pays that for a total nothing reads back.

The completion-detection point stands, but it only applies when the caller opts into waiting, which is their explicit choice and already the documented trade on insert. It should not force the count on someone who never asked to wait.

All three delete paths now gate the count the same way the insert paths do: `python.py:_delete_dataframe_kafka`, `spark.py:_delete_online_dataframe`, and `SparkEngine.deleteOnlineDataframe`. `remove_rows` gained the `write_options` documentation it lacked, including that disabling the count means `wait_for_online_ingestion` runs to `timeout`. Unit tests cover the gate on the Python and Spark engines.

One thing this surfaced, pre-existing and not limited to deletes: disabling the count while setting `timeout` to `0` leaves neither break in `wait_for_completion` reachable (`online_ingestion.py:207,211`, and `OnlineIngestion.java:83,88`), so the wait hangs rather than timing out. Noted on every docstring carrying the option; worth a proper fix separately.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/FeatureGroupEngine.java
```diff
@@ -468,14 +468,48 @@ public Map<Long, Map<String, String>> commitDetailsByWallclockTime(FeatureGroupB
...
+    }
+
+    // Offline delete is always applied; online delete is opt-in and routed as a Kafka
+    // tombstone consumed by OnlineFS, which deletes the row from RonDB by primary key.
+    FeatureGroupCommit commit = HudiEngine.getInstance().deleteRecord(
+        SparkEngine.getInstance().getSparkSession(), featureGroupBase, genericDataset, writeOptions);
+
+    if (deleteOnline) {
+      if (featureGroupBase instanceof StreamFeatureGroup) {
+        FeatureGroupEngineBase.LOGGER.warn(
```
bubriks: why? cant we update delta streamer both for delta/iceberg and hudi? we already deal with headers within them.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/SparkEngine.java
```diff
@@ -584,10 +587,69 @@ public <S> StreamingQuery writeStreamDataframe(FeatureGroupBase featureGroupBase
...
+   * <p>Each message carries the row encoded against the feature group Avro schema plus an
+   * {@code operation: delete} header; OnlineFS deletes the row by primary key and ignores the values.
+   * The dataset needs to carry only the primary key: non-key columns are filled with null so the Avro
+   * schema serializes.
+   *
+   * @param featureGroupBase the online-enabled feature group
+   * @param dataset the rows to delete
+   * @param writeOptions kafka write options
+   */
+  public void deleteOnlineDataframe(FeatureGroupBase featureGroupBase, Dataset<Row> dataset,
```
bubriks, line 605: not sure but maybe we have to look into UI/api as there these deletes might show up as upserted not deleted. (it might confuse user) the three statuses in online ingestion result i think are Upserted, Ignored, Failed.

python/hsfs/engine/spark.py
```diff
@@ -832,17 +832,122 @@ def _save_online_dataframe(self, feature_group, dataframe, write_options):
...
+        for field in json.loads(feature_group.avro_schema)["fields"]:
+            if field["name"] in primary_key:
+                continue
+            dataframe = dataframe.withColumn(
+                field["name"], lit(None).cast(self._avro_to_spark_type(field["type"]))
+            )
+        return dataframe
+
+    @staticmethod
+    def _avro_to_spark_type(avro_type):
```
bubriks: why do we need this? do we have to cast None, cant it be done another way?

bubriks (changes requested):

python/hsfs/feature_group.py
```diff
@@ -4695,39 +4697,113 @@ def commit_details(
         return self._feature_group_engine._commit_details(self, wallclock_time, limit)
 
     @public
-    def commit_delete_record(
+    def remove_rows(
         self,
         delete_df: TypeVar("pyspark.sql.DataFrame"),  # noqa: F821
         write_options: dict[Any, Any] | None = None,
+        delete_online: bool = False,
```
manu-sj: I think this should be `True` by default to maintain consistency while deleting.

We keep it as `False` in `commit_delete_record` to maintain backwards compatibility.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/SparkEngine.java
dhananjay-mk: added deleted status in backend and ui

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/FeatureGroupEngine.java
dhananjay-mk: add support for stream fg

python/hsfs/engine/spark.py
dhananjay-mk: updated

python/hsfs/feature_group.py
```diff
@@ -4695,39 +4697,115 @@ def commit_details(
         return self._feature_group_engine._commit_details(self, wallclock_time, limit)
 
     @public
-    def commit_delete_record(
+    def remove_rows(
         self,
         delete_df: TypeVar("pyspark.sql.DataFrame"),  # noqa: F821
         write_options: dict[Any, Any] | None = None,
+        delete_online: bool = True,
     ) -> None:
```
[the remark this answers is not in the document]
dhananjay-mk: Fixed in 4c0622ef5, taking a third option rather than either of the two suggested.

`delete_online` is now tri-state (`bool | None = None`). Left unset it follows the feature group, so `remove_rows(delete_df)` deletes from both stores on an online-enabled feature group and offline-only on one without an online store. Passing `delete_online=True` explicitly on a feature group that has no online store still raises, so option (b)'s silent no-op does not apply to a request the caller actually made, only to the unspecified case.

The same defect applied to feature groups with an embedding index, whose online data lives in the vector database: with the default resolving to `True` those raised too. The auto path now treats "has an embedding index" as "cannot delete online" and goes offline-only, while an explicit `True` still raises.

Docstring, the `hops-fg` skill and the PR description are updated to match, and the tests cover the unset resolution plus both explicit-raise paths.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/FeatureGroup.java
```diff
@@ -883,19 +883,22 @@ public Object insertStream(Dataset<Row> featureData, String queryName, String ou
...
    * @throws FeatureStoreException If Client is not connected to Hopsworks and/or no commit information was found for
    *                               this feature group;
    * @throws IOException Generic IO exception.
    * @throws ParseException In case it's unable to parse HUDI commit date string to date type.
    */
-  public void commitDeleteRecord(Dataset<Row> featureData)
+  public void removeRows(Dataset<Row> featureData)
       throws FeatureStoreException, IOException, ParseException {
-    featureGroupEngine.commitDelete(this, featureData, null);
+    featureGroupEngine.commitDelete(this, featureData, null, true);
```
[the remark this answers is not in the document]
dhananjay-mk: Fixed in 4c0622ef5. Both overloads that take no `deleteOnline` argument (this one and line 937) now pass `Boolean.TRUE.equals(getOnlineEnabled())` instead of a hardcoded `true`, so they follow the feature group the same way the tri-state Python default does.

`TestFeatureGroup.testRemoveRowsFollowsOnlineEnabled` asserts the flag that reaches `commitDelete` for both an online-enabled and an offline-only feature group. To make that assertion possible without reflection, `FeatureGroup.featureGroupEngine` dropped from `private final` to `protected`, matching the identical field on `StreamFeatureGroup`.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/StreamFeatureGroup.java
```diff
@@ -1107,19 +1107,22 @@ public Query selectExcept(List<String> features) {
...
    * @throws FeatureStoreException If Client is not connected to Hopsworks and/or no commit information was found for
    *                               this feature group;
    * @throws IOException Generic IO exception.
    * @throws ParseException In case it's unable to parse HUDI commit date string to date type.
    */
-  public void commitDeleteRecord(Dataset<Row>  featureData)
+  public void removeRows(Dataset<Row>  featureData)
       throws FeatureStoreException, IOException, ParseException {
-    featureGroupEngine.commitDelete(this, featureData, null);
+    featureGroupEngine.commitDelete(this, featureData, null, true);
```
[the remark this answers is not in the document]
dhananjay-mk: Fixed in 4c0622ef5. Same change as on `FeatureGroup`: this overload and the one at line 1160 now pass `Boolean.TRUE.equals(getOnlineEnabled())` rather than a hardcoded `true`, so a stream feature group without an online store is deleted from offline only instead of hitting the `commitDelete` guard.

Covered by `TestFeatureGroup.testStreamFeatureGroupRemoveRowsFollowsOnlineEnabled`.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/hudi/DeltaStreamerAvroDeserializer.java
```diff
@@ -106,6 +106,12 @@ public void configure(Map<String, ?> configs, boolean isKey) {
   public GenericRecord deserialize(String topic, Headers headers, byte[] data) {
     if (subjectId.equals(getHeader(headers, "subjectId"))
         && featureGroupId.equals(getHeader(headers, "featureGroupId"))) {
+      // A delete tombstone is on the topic for OnlineFS only: removeRows has already applied the
+      // offline delete directly to the table, so materializing the tombstone re-inserts the key
+      // it deleted. Exact match, as in OnlineFsHandler.getRow - any other value is an upsert.
+      if ("delete".equals(getHeader(headers, "operation"))) {
```
bubriks: maybe instead we could rely on `storage` header. i created it couple months ago, but it seams i didnt add changes here and in utils for materialization job.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/FeatureGroup.java
```diff
@@ -915,20 +920,136 @@ public void commitDeleteRecord(Dataset<Row> featureData)
...
+   * passed for a non-key feature has no effect on it.
+   *
+   * @param featureData Spark DataFrame, RDD. Feature data to be deleted.
+   * @param deleteOnline Also delete the records from the online store when the feature group is online-enabled.
+   * @throws FeatureStoreException If deleteOnline is set and the feature group is not online-enabled;
+   *                               or on other client/commit errors.
+   * @throws IOException Generic IO exception.
+   * @throws ParseException In case it's unable to parse HUDI commit date string to date type.
+   */
+  public void removeRows(Dataset<Row> featureData, boolean deleteOnline)
```
bubriks: i think it should be like insert, use `Storage`
`public void insert(Dataset<Row> featureData, Storage storage)`

python/hsfs/engine/python.py
```diff
@@ -1981,6 +1981,91 @@ def _write_dataframe_kafka(
                 options=offline_write_options.get("online_ingestion_options", {})
             )
 
+    def _delete_dataframe_kafka(
```
bubriks, line 2013: couldn't we reuse `_write_dataframe_kafka` at a glance it looks pretty much the same?

python/hsfs/engine/spark.py
```diff
@@ -832,17 +832,78 @@ def _save_online_dataframe(self, feature_group, dataframe, write_options):
                 options=write_options.get("online_ingestion_options", {})
             )
 
+    def _delete_online_dataframe(self, feature_group, dataframe, write_options):
```
bubriks, line 844: same question as in python, can we reuse `_save_online_dataframe`?

python/hsfs/feature_group.py
```diff
@@ -4695,39 +4697,126 @@ def commit_details(
         return self._feature_group_engine._commit_details(self, wallclock_time, limit)
 
     @public
-    def commit_delete_record(
+    def remove_rows(
         self,
         delete_df: TypeVar("pyspark.sql.DataFrame"),  # noqa: F821
         write_options: dict[Any, Any] | None = None,
+        delete_online: bool | None = None,
```
bubriks: use storage like for insert methods.

bubriks (changes requested):

python/hsfs/feature_group.py
dhananjay-mk: Done, on both clients. `commitDelete` now owns the null-follows-the-feature-group resolution that both FG classes duplicated.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/FeatureGroup.java
dhananjay-mk: Covered in the reply on the Python thread: `removeRows(Dataset, Storage)` and `removeRows(Dataset, Map, Storage)` on both `FeatureGroup` and `StreamFeatureGroup`, with `Storage.ONLINE` reserved and raising for now.

python/hsfs/engine/spark.py
dhananjay-mk: Done. Merged into `_save_online_dataframe(..., operation=)`; `_delete_online_dataframe` is now a two-line delegator. They differed only in the dataframe prep step and the header.

python/hsfs/engine/python.py
dhananjay-mk: Partly. The shared scaffolding is extracted (`_iter_dataframe_rows`, `_produce_row_kafka`, `_wait_for_online_ingestion`) and both paths use it.

I stopped short of folding delete into `_write_dataframe_kafka` itself. That method carries `_mark_online_rows`, the per-row storage header, the online-only row skip, the n_rows-from-flags branch and the `_multi_part_insert` flush, none of which a delete wants, so it would mean five or six new branches in the insert hot path.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/hudi/DeltaStreamerAvroDeserializer.java
dhananjay-mk: You are right that the branch reads wrong: it checks an operation to decide a destination, and needed three lines of comment to bridge that. Extracted behind `skipOfflineMaterialization(headers)` so the call site reads as the routing decision, with the reasoning inside the method.

I kept `operation` as the check, though. The remaining path for offline-delete-through-the-stream is a custom payload keyed on the operation header (the `_hoodie_is_deleted` marker route is dead: persisting the column fails every later client write with `MissingSchemaFieldException`). So `operation` stays load-bearing there, and that future change is one line in this method. With a routing header the tombstone would also have to stop carrying it, or we drop the record before the operation branch ever runs.

Separately, you are right that the axis lacks an "online only" value, and adding one is purely additive since any non-`"0"` value is already ingested. Happy to file it as its own ticket. Which utils did you have in mind on the materialization side? The deserializer is the only place I could find that reads headers.

python/hsfs/feature_group.py
dhananjay-mk: Correction to my earlier reply on this thread: `storage="online"` no longer raises. It does what the name says, and what the same value means on `insert`: the online delete runs and the offline commit is skipped.

I had reserved it because a single-store delete is the one mode nothing reconciles afterwards, and wanted that to be a deliberate product call rather than a side effect of the enum having three values. Having raised it, the answer was that there is no technical reason to withhold it, so it is implemented.

Two things came out of doing it that are worth knowing:

The HUDI and ICEBERG engine guards in `remove_rows` were unconditional, but both are properties of the *offline* delete: HUDI needs Spark, ICEBERG needs pyiceberg without it. They now run only when the offline leg does, so an online-only delete is reachable on feature groups where `remove_rows` previously could not run at all.

Online-only raises on a feature group with no online store and on one with an embedding index. Both would delete nothing, and with the offline leg skipped there is no partial result worth keeping, unlike the unset case where an embedding index warns and the offline delete still lands.

`commit_delete_record` also went back to its exact upstream signature. This branch had added a `delete_online` argument to it; that is redundant now that `remove_rows` takes `storage`, and growing a deprecated method works against the deprecation. It is unreleased, so nothing breaks. Java lost the two boolean `commitDeleteRecord` overloads for the same reason.

Verified on a cluster: python driver 7 passed including a new online-only case that asserts the row is gone from the online store and still present offline, pyspark driver 4 passed across stream DELTA, HUDI and ICEBERG.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/FeatureGroup.java
dhananjay-mk: Correction to my earlier reply: `Storage.ONLINE` is no longer reserved. It deletes from the online store and skips the offline commit, which is what the same value means on `insert`. It raises only where an online-only delete would remove nothing, so on a feature group with no online store and on one with an embedding index.

`commitDeleteRecord` is back at its exact upstream signature. The two boolean overloads this branch had added are gone, since `removeRows(Dataset, Storage)` expresses the same choice and growing a deprecated method works against the deprecation.

python/hsfs/feature_group.py
dhananjay-mk: `delete_online` is gone. `remove_rows` takes `storage` instead, matching `insert`: unset deletes from both stores, `"offline"` and `"online"` each pick one.

That gives the default you asked for without the boolean. Unset deletes online whenever the feature group is online-enabled. `commit_delete_record` keeps its backward-compatible behaviour by delegating with `storage="offline"`, and is back at its exact upstream signature.

bubriks (approved):

PR #1074, opened 2026-07-28
[HWORKS-2981] Treat stranded QUEUE_TIMEOUT executions as terminal job failures

javierdlrm:
https://hopsworks.atlassian.net/browse/HWORKS-2981

Adds `QUEUE_TIMEOUT` and `SUBMISSION_FAILED` to the Python SDK's job `ERROR_STATES` and `COMPLETED` to `SUCCESS_STATES`, so `await_termination` raises instead of polling forever on executions stranded in those states.

The backend normally persists `state=QUEUE_TIMEOUT` together with `final_status=FAILED` (`KubeJobsMonitor.updateFailedState`), which `ERROR_STATES` already catches through `final_status`. The hazard is the crash window: `updateState` and `updateFinalStatus` are two separate facade calls, and if the jobs monitor dies between them the row stays at `state=QUEUE_TIMEOUT, final_status=UNDEFINED` forever. `QUEUE_TIMEOUT` is not in the server's `JobState.getFinalStates()`, so `ExecutionFacade.findExpiredFinished` never reconciles it, and a client polling `await_termination` hangs indefinitely.

`SUBMISSION_FAILED` and `COMPLETED` are defensive: the Spark and Ray monitors currently rewrite `SUBMISSION_FAILED` to `FAILED` and normalize `COMPLETED` to `FINISHED` before persisting, but both are one persistence change away from reproducing the identical hang.

Tests cover both the stranded row (`final_status="UNDEFINED"`, fails without this change) and the state the backend actually writes (`final_status="FAILED"`, green before and after, pinning the production path).

Split out of #1019 per review (unrelated to KServe; original commit message misattributed the root cause).

### DIFF
--- a/python/hopsworks_common/constants.py
+++ b/python/hopsworks_common/constants.py
@@ -32,13 +32,18 @@ DEFAULT = Default()  # TODO: figure out what to do with it
 
 
 class JOBS:
-    SUCCESS_STATES = ["FINISHED", "SUCCEEDED"]
+    # COMPLETED is defensive: the monitors currently normalize it to FINISHED
+    # before persisting, but it is in the server's final-states set and would
+    # otherwise reproduce the same await_termination hang if ever persisted.
+    SUCCESS_STATES = ["FINISHED", "SUCCEEDED", "COMPLETED"]
     ERROR_STATES = [
         "FAILED",
         "KILLED",
         "FRAMEWORK_FAILURE",
         "APP_MASTER_START_FAILED",
         "INITIALIZATION_FAILED",
+        "SUBMISSION_FAILED",
+        "QUEUE_TIMEOUT",
     ]
 
 
--- a/python/tests/core/test_job.py
+++ b/python/tests/core/test_job.py
@@ -420,6 +420,89 @@ class TestJob:
         # Assert
         assert str(e_info.value) == "The Hopsworks Job was stopped"
 
+    def test_run_await_termination_python_raises_on_queue_timeout(
+        self, mocker, backend_fixtures
+    ):
+        # Arrange — the backend normally persists QUEUE_TIMEOUT together with
+        # final_status=FAILED, but updateState and updateFinalStatus are two
+        # separate facade calls: if the jobs monitor dies between them the row
+        # is stranded at state=QUEUE_TIMEOUT, final_status=UNDEFINED forever
+        # (QUEUE_TIMEOUT is not in the server's final-states set, so nothing
+        # reconciles it). await_termination must raise on state alone, not hang.
+        mocker.patch("hopsworks_common.client._get_instance")
+        mocker.patch("hopsworks_common.execution.Execution.get_url")
+        mock_execution_api = mocker.patch(
+            "hopsworks_common.core.execution_api.ExecutionApi",
+        )
+        python_job_mock = mocker.Mock()
+        python_job_mock.job_type = "PYTHON"
+        mock_execution_api.return_value._start.return_value = execution.Execution(
+            job=python_job_mock
+        )
+        mock_execution_api.return_value._get.return_value = execution.Execution(
+            id=1, state="QUEUE_TIMEOUT", final_status="UNDEFINED", job=python_job_mock
+        )
+
+        j = job.Job(
+            id="test_id",
+            name="test_name",
+            creation_time=None,
+            config={},
+            job_type="PYTHON",
+            creator=None,
+        )
+
+        # Act
+        with pytest.raises(exceptions.JobExecutionException) as e_info:
+            j.run(await_termination=True)
+
+        # Assert
+        assert (
+            str(e_info.value)
+            == "Execution failed with status: QUEUE_TIMEOUT. See the logs for more information."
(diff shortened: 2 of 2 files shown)


robzor92 (commented):
Checked the premise against `hopsworks-ee@master` and it holds up: `KubeJobsMonitor` is `@Singleton @TransactionAttribute(NOT_SUPPORTED)` and `ExecutionUpdateController` is `@Stateless @TransactionAttribute(NOT_SUPPORTED)`, so `updateFailedState`'s `updateState()` and `updateFinalStatus()` really do commit in separate transactions — the stranded row is reachable, not hypothetical. And nothing repairs it, because `QUEUE_TIMEOUT ∉ JobState.getKubeRunningStates()` so `findByTypesAndStates` never re-selects the execution.

Two things I want to call out as *right*, since they're the subtle part: the resulting set is exactly `JobState.getFinalStates() ∪ {QUEUE_TIMEOUT}`, and the PR adds `QUEUE_TIMEOUT` **without** adding `QUEUED`/`PREEMPTED`. That second point is what makes this safe on Kueue clusters — a queued or preempted workload is legitimately long-lived and gets re-admitted, so marking either terminal would have turned normal quota waiting into a client-side failure. `QUEUE_TIMEOUT` itself is not Kueue-only (`checkPodWaitingResources()` sets `QUEUED` with no Kueue gate, and the DLT multi-table path emits `QUEUE_TIMEOUT` *only* when Kueue is off), so the fix is needed on both cluster shapes and behaves identically on both.

Four things I'd fix before merge.

### 1. The new stranded-row test hangs instead of failing when the fix is reverted

I reverted `constants.py` on the PR head and ran `pytest tests/core/test_job.py -k queue_timeout`: it had to be killed at 90 s. It never fails — `_get` returns `QUEUE_TIMEOUT`/`UNDEFINED` forever, `job.run(await_termination=True)` passes no timeout into `_wait_until_finished`, and `pytest-timeout` isn't in the dev deps. So if this regresses, CI gives you a hung runner up to the 6 h default, not a red X.

(The second test, `final_status="FAILED"`, passes with and without the change — matching the description.)

Cheapest fix: assert directly that `Execution(state="QUEUE_TIMEOUT", final_status="UNDEFINED", …).success is False`, and keep one bounded end-to-end test.

### 2. `app.py` holds a second mirror of the same set, and it wasn't updated

`hopsworks_common/app.py:35-48` — the comment literally says *"mirrors `JobState.getFinalStates()`"* — has `SUBMISSION_FAILED` but not `QUEUE_TIMEOUT`. `_wait_for_serving` polls it at `app.py:515`, so a queue-timed-out app burns the full 600 s `SERVING_TIMEOUT` and then reports *"Timed out waiting for app to reach Serving state"* instead of the actual cause.

It's reachable: `PYTHON_APP ∈ KubeJobsMonitor.MONITORED_JOB_TYPES`, and the pending-execution branch that writes `QUEUE_TIMEOUT` is job-type agnostic. One line.

### 3. Public typing still says this state can't happen

`Job.get_state()`'s `Literal[...]` (`job.py:279-297`) and the `Execution.state` docstring (`execution.py:106-109`) both omit `QUEUE_TIMEOUT`. The point of this PR is that users will now see that string in the exception message, and with *Unit Tests (Typechecked)* running in CI the `Literal` is a lie they can trip over. Adding it to `get_state()` at minimum.

Separately (pre-existing, not for this PR): `get_final_state()`'s `Literal` lists `FRAMEWORK_FAILURE`/`APP_MASTER_START_FAILED`/`INITIALIZATION_FAILED`, but `JobFinalStatus` only has `UNDEFINED|SUCCEEDED|FAILED|KILLED` — wrong in the other direction.

### 4. The `findExpiredFinished` claim in the description/commit message is wrong

The named query is

```sql
SELECT e FROM Execution e WHERE e.state IN :finalStates AND e.executionStop > 0 AND e.executionStop < :cutoff
```

and its only caller is `ExecutionsCleaner.deleteOrphanExecutions`. It's the TTL **garbage collector**, not a reconciler — it would never have repaired `final_status`, so it can't be the reason the client hangs. The accurate reason is the `getKubeRunningStates()` one above.

What `QUEUE_TIMEOUT ∉ getFinalStates()` *actually* costs is that these rows are **never TTL-deleted** and accumulate indefinitely. Worth correcting, since this paragraph is the justification a future reader will lean on.

---

Smaller notes: Ray never produces `COMPLETED` (`RayJobsMonitor.getState()` normalizes `SUCCEEDED`→`FINISHED`) — only Spark does the `COMPLETED`→`FINISHED` rewrite, so the "Spark and Ray … normalize `COMPLETED`" wording is a bit loose. And the `constants.py` comment explains `COMPLETED`, where the rationale is weakest, but not the two `ERROR_STATES` additions, where it's load-bearing.

I also have three backend follow-ups out of this (a 2-line ordering fix in `updateFailedState` that would close the whole class of stranded rows, a Kueue workload-release gap on the `QUEUE_TIMEOUT` path, and the same missing states in the ruby IT helper) — happy to file those separately.

PR #1090, opened 2026-08-10
[FSTORE-2068] Backport online delete on remove_rows to 5.0

dhananjay-mk:
Backport of #1069 to `branch-5.0`: cherry-pick of `03f5c6fba`, plus the doc fix from #1100 picked from main (`f8b31da77`).

## Deliberate differences from main

All three because `branch-5.0` has no ICEBERG feature groups (the backend enum is `NONE, HUDI, DELTA`):

- the ICEBERG guard in `remove_rows` is dropped, since `HAS_PYICEBERG` does not exist here and the branch would be unreachable
- the ICEBERG param on the stream delete test is dropped; that left the test identical to `test_online_delete_by_default`, so the duplicate went too
- the `hops-fg` skill's "DELTA, ICEBERG, and HUDI" line now reads "both DELTA and HUDI"

The HUDI guard is unchanged.

## Conflicts

Two, both in `python/hsfs/feature_group.py`: the apigen import line gained `deprecated`, and `commit_delete_record`'s body is replaced by the new implementation. Everything else auto-merged.

## Verification

`ruff check` and `ruff format --check` clean. `pytest python/tests`: 3459 passed, 13 skipped, 5 failed. The 5 failures are `TypeError: 'JavaPackage' object is not callable` in `test_spark.py` avro tests and fail identically on a pristine `branch-5.0` worktree (missing local spark-avro jar), so they are the baseline, not a regression.

No cluster validation: the dev cluster runs 5.1-line images, so it would prove nothing about the 5.0 artifacts.

## Companion backports

clusterj-onlinefs #182, hopsworks-ee #3208, hopsworks-front #2031, loadtest #972 (merges last, needs a 5.0.x SDK release first).

[FSTORE-2068]: https://hopsworks.atlassian.net/browse/FSTORE-2068

### DIFF
--- a/python/hsfs/feature_group.py
+++ b/python/hsfs/feature_group.py
@@ -32,7 +32,7 @@ from urllib.parse import urlparse
 import avro.schema
 import hsfs.expectation_suite
 import humps
-from hopsworks_apigen import deprecation, public
+from hopsworks_apigen import deprecated, deprecation, public
 from hopsworks_common import job
 from hopsworks_common.client.exceptions import FeatureStoreException, RestAPIError
 from hopsworks_common.core import alerts_api
@@ -3692,6 +3692,7 @@ class FeatureGroup(FeatureGroupBase):
                     - `mark_online_rows`: `True` or `False` to filter rows for online ingestion based on event time and primary key deduplication, defaults to `True`.
                     - `disable_online_ingestion_count`: `True` or `False` to disable sending the total number of entries to the online ingestion tracking system, defaults to `False`.
                       When `True`, no batch size is known to the ingestion tracker so `wait_for_online_ingestion` will wait until `timeout` is reached.
+                      Combining it with `timeout` `0` waits forever, since neither the entry count nor the timeout can end the wait.
                 - key `start_offline_backfill` and value `True` or `False` to configure whether or not to start the materialization job to write data to the offline storage. `start_offline_backfill` is deprecated.
                   Use `start_offline_materialization` instead.
                 - key `start_offline_materialization` and value `True` or `False` to configure whether or not to start the materialization job to write data to the offline storage.
@@ -3933,6 +3934,7 @@ class FeatureGroup(FeatureGroupBase):
                     - `mark_online_rows`: `True` or `False` to filter rows for online ingestion based on event time and primary key deduplication, defaults to `True`.
                     - `disable_online_ingestion_count`: `True` or `False` to disable sending the total number of entries to the online ingestion tracking system, defaults to `False`.
                       When `True`, no batch size is known to the ingestion tracker so `wait_for_online_ingestion` will wait until `timeout` is reached.
+                      Combining it with `timeout` `0` waits forever, since neither the entry count nor the timeout can end the wait.
                 - key `start_offline_backfill` and value `True` or `False` to configure whether or not to start the materialization job to write data to the offline storage.
                   `start_offline_backfill` is deprecated.
                   Use `start_offline_materialization` instead.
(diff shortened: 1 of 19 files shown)


python/tests/engine/test_python_delete.py
```diff
@@ -0,0 +1,356 @@
...
+        time_travel_format=time_travel_format,
+    )
+    fg.primary_key = ["id"]
+    return fg
+
+
+class TestRemoveRowsStreamFeatureGroup:
+    # ICEBERG alongside DELTA because _resolve_stream_python makes stream=True the default
+    # for ICEBERG on external clients, so it is the common stream case rather than an edge one.
+    @pytest.mark.parametrize("time_travel_format", ["DELTA", "ICEBERG"])
```
manu-sj: Remove iceberg

dhananjay-mk: done

python/hsfs/feature_group.py
```diff
@@ -4430,29 +4338,149 @@ def commit_details(
...
+                "The offline delete was applied, so the two stores now differ for the removed rows.",
+                stacklevel=2,
+            )
+            delete_online = False
+
+        if delete_offline:
+            # The guard is a property of the offline delete: a HUDI delete needs Spark.
+            # An online-only delete goes to Kafka and needs neither, so it stays available
+            # on those feature groups.
+            # The 5.1 line also guards ICEBERG here. This release has no ICEBERG feature
```
manu-sj: Maybe we should remove comments for iceberg here too

skills/hops/hops-fg/SKILL.md
```diff
@@ -478,11 +482,34 @@ rows_to_delete = pl.DataFrame({
...
+```python
+fg.remove_rows(rows_to_delete)
+```
+
+Pass `delete_online=False` to leave the online store untouched and delete offline only.
+
+The online delete matches on the primary key alone, so every non-key column in the DataFrame is ignored for the online leg, `event_time` included.
+The key columns the offline merge requires are still required; they just do not affect which online rows are removed.
+
+Stream FGs are included: the online delete works for all of DELTA, ICEBERG, and HUDI.
```
manu-sj: ICEBERG not in 5.0, so we maybe remove it from here too

python/hsfs/feature_group.py
dhananjay-mk: Dropped in 207d0a944.

I also reworded the line just above it. It said an online-only delete "needs neither", which was counting Spark and pyiceberg, so with the ICEBERG guard gone the "neither" had nothing left to refer to. It now reads "goes to Kafka and does not". The HUDI guard itself is unchanged.

skills/hops/hops-fg/SKILL.md
dhananjay-mk: Fixed in 207d0a944. The line now reads "the online delete works for both DELTA and HUDI".

Checking that line turned up a separate problem in the same section: it documented a `delete_online` keyword in four places, including the cheatsheet row `fg.remove_rows(df, delete_online=True)`. That keyword does not exist. `remove_rows` takes `storage`, so anyone copying the cheatsheet row gets a TypeError. #1069 swapped the boolean for `storage` during review and the skill kept the old spelling, which means it was wrong on main as well, not only here.

Since it was a live bug on both branches I raised it on main first as #1100 (now merged) and cherry-picked it here as 61b5b1ccc, so the two branches stay convergent on this section. That commit also corrects two claims that came with the old boolean: `commit_delete_record` has no parameter reaching the online store at all, and an embedding-index feature group warns and continues on a default delete, raising only under `storage="online"`.

manu-sj (approved):

PR #1104, opened 2026-08-17
[FSTORE-2096] Attachment timestamps, the keyword API, and tags for jobs, apps and deployments

jimdowling:
Ticket: [FSTORE-2096](https://hopsworks.atlassian.net/browse/FSTORE-2096) · Backend: logicalclocks/hopsworks-ee#3227 · Frontend: logicalclocks/hopsworks-front#2042 · Helm: logicalclocks/hopsworks-helm#2212 · Docs: logicalclocks/logicalclocks.github.io#635

The Python client half of the FSTORE-2075 split (api#1079): the low-risk, independently useful parts, landing first. 21 files, ~2,100 added lines.

## Attachment timestamps

`Tag` objects carry `created_on`, parsed tolerantly from epoch millis, ISO strings, or absence, because pre-existing rows report NULL by design. Feature groups, feature views, and training datasets gain `get_tag_metadata` / `get_tags_metadata` alongside the value-only reads, and keyword metadata reads report each keyword with its attachment time.

## The keyword API

A real `keywords_api` addressing the keyword endpoints, including the cluster-wide vocabulary read on the feature store. The CLI keyword commands on `fg`, `fv`, and `td` operate on keywords rather than on tags, which is what they were mislabelled as doing; the retired `add-keyword --value` option warns loudly on stderr, where a click command's user actually sees it.

## Job tags and a job description

`Job` objects gain tag reads and writes (`add_tag`, `get_tag(s)`, `get_tag(s)_metadata`, `delete_tag`) and a `description` property that answers config-first, so an older server that never sent the scalar field still reports the description its configuration carries. The setter writes both, guarded the same way the getter is.

## A pre-existing defect fixed on the way

`job.py` defined the `config` property **twice**, and the second definition shadowed the first's getter+setter pair, so `job.config = x` raised `AttributeError: property 'config' of 'Job' object has no setter`. Found by the moved tests; the duplicate is gone and a test pins the setter.

## What deliberately stays with FSTORE-2075 (api#1079)

The search API and its result types, the tag schema lifecycle client (`deprecate`/`restore`/`usage`/guarded delete and the `hops tags` CLI group), dataset tags, and the foreign-`Project` handle binding. After this merges, api#1079 rebases and shrinks by this amount.

## Verification

Full unit suite: **4145 passed, 13 skipped, 0 failed** (with `PYSPARK_SUBMIT_ARGS` per CI). ruff check and format clean, docsig clean under the project config, the `@public` gate passes. New coverage: `TestJobDescription` and `TestJobTags` (12 tests), keyword API tests, tag `created_on` parsing tests.

Synced with `upstream/main` (through `[FSTORE-2086]` #1105); no conflicts and no client change was needed for it.

**Cluster-verified on jim-tags** alongside ee#3227's EAR: job tag attach returns 201 with `createdOn`, keywords answer JSON with no `Accept` header and carry per-keyword `createdOn`, and feature-group / model / deployment tags report `created_on` with a re-attach preserving the original time. The search-side work this ticket added since (apps and agents as their own result classes, the tag-facets endpoint) is server plus UI only and adds no client surface, so nothing here changed for it.

## Loadtests

New coverage: **logicalclocks/loadtest#984**, 13 tests, **13/13 passing** on jim-tags against this EAR. Job tags (attach/read/delete, `createdOn` reported, re-attach preserving the original `createdOn`, cascade on job delete), the `JOB`/`APP` partition in both directions with `ALL` double-counting nothing, tag content matching a job document, tag facets reporting the attached values, and the `archive` flag defaulting false / settable / per schema.

Existing suite, the six files that cover what this touches (`hopsworks/test_search.py`, `hopsworks/test_opensearch.py`, `hopsworks/test_custom_apps.py`, `hopsworks/test_python_job_execution.py`, `feature_store/python_driver/test_tags.py`, `model_serving/python_driver/external/test_tags.py`): **21 collected, 15 passed, 1 failed, 5 skipped**.

The one failure is environmental, not a regression: `test_custom_app_lifecycle[fastapi]` timed out with `JobExecutionException: Timed out waiting for app to reach Serving state after 600.0s. Current state: RUNNING` — the app was created and reached RUNNING but never got the resources to serve. The 5 skips are its siblings, skipped after it failed. Nothing in this PR touches app startup or the serving state machine.

A full-suite run was attempted first and abandoned. On a two-worker cluster it was capacity-bound rather than slow: a Spark driver sat `Pending` for 78 minutes, and after freeing capacity the rate settled at ~12 min/test, a ~5-day projection for 935 tests. Its only failure was also environmental (`test_pyspark_batch[DELTA]`, executor starvation then a 600 s shuffle-service idle timeout, no application error). A run that manufactures failures at that rate cannot serve as evidence, which is why the targeted subset is what is reported here.

**Two real defects the runs found, both now fixed:**

- **A deployment-state gap.** `POST /tags` returned `500 {"errorCode":120005,"usrMsg":"Unknown column 'archive' in 'field list'"}` on a cluster where V92 had been applied *before* the `archive` column was appended to it. Flyway will not re-run a recorded migration, so the EAR had the entity field while the table did not. A fresh install is unaffected; anyone who ran a pre-release V92 is not. Applied on the cluster and verified: create without the flag reads back `archive=False`, with `archive=true` reads back `True`.
- **The kube runner ran only the first path given to it.** `run_tests_on_kube.sh` passed `${1:-workflows/*}` to pytest while `deploy_loadtest.sh` forwards any number of paths, so a six-file selection ran one file and exited 0 — reading as "your selection passed" when most of it never ran. Collection went from 4 items to 21 once fixed. Fixed in loadtest#984.

[FSTORE-2096]: https://hopsworks.atlassian.net/browse/FSTORE-2096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[FSTORE-2086]: https://hopsworks.atlassian.net/browse/FSTORE-2086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/hopsworks/cli/commands/fg.py
+++ b/python/hopsworks/cli/commands/fg.py
@@ -3,7 +3,8 @@
 Covers the full feature-group surface: ``list``, ``info``, ``preview``,
 ``features`` (reads) plus ``create``, ``create-external``, ``insert``,
 ``derive``, ``append-features``, ``delete``, ``stats``, ``search``,
-``keywords``/``add-keyword``/``remove-keyword`` (writes). All operations go
+``keywords``/``add-keyword``/``remove-keyword`` (plain labels) and
+``tags``/``add-tag``/``remove-tag`` (name/value pairs). All operations go
 through the SDK in-process so Hopsworks domain logic is invoked directly, with
 no subprocess hop.
 """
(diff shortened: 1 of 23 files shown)


o-alex:
**Loadtest coverage before this leaves draft**

The PR body says cluster verification is "to follow together with ee#3227". For a Python-client change the integration run against a cluster is the only place the SDK is actually exercised, so it is the precondition for being out of draft rather than a follow-up.

The tests for it already exist on FSTORE-2075's loadtest branch: `workflows/hopsworks/test_job_metadata.py` (+ `test_code/job_metadata.py`) covers the job description round trip, the length refusal, job tags and their cascade on delete, and `workflows/feature_store/test_code/tags.py` carries `test_tag_created_on`, `test_keyword_round_trip` and `test_keyword_created_on` — all of which are this ticket's surface, not the schema lifecycle's.

Details and the full cherry-pick list are in https://github.com/logicalclocks/hopsworks-ee/pull/3227#issuecomment-5341751469. Cherry-picking those into a sibling loadtest PR under FSTORE-2096 and running it would cover this PR too, and would let the `test_job_metadata.py` 422→400 assertion be settled in the ticket that introduces the typed code.

o-alex (approved):

PR #1105, opened 2026-08-18
[FSTORE-2086] Add support for disabling Uploads to Hopsworks

manu-sj:
Part of FSTORE-2086, and the fix for the blocker raised on logicalclocks/hopsworks-ee#3224.

Related PRs: hopsworks-ee#3224 (backend enforcement), hopsworks-helm#2201 (chart value), hopsworks-front#2041 (UI), logicalclocks.github.io#632 (docs).

## The bug

`dataset_api.upload(..., overwrite=True)` is implemented as delete-then-upload, because the dataset upload endpoint cannot overwrite:

```python
if self.exists(destination_path):
    if overwrite:
        ...
        self.remove(destination_path)      # DELETE succeeds: not policy-gated
...
self._upload_file(...)                     # POST -> 403 under the new upload policy
```

FSTORE-2086 gives the backend a cluster upload policy (`enabled` / `admins_only` / `disabled`) enforced on the upload endpoint. `DatasetResource.delete` is deliberately not gated by it. So on a cluster with uploads restricted, the removal lands and the upload that follows is refused. **The destination is gone and there is nothing left to restore.**

For a directory the removal is recursive, so it also takes files that exist only on the server and have no local counterpart. There is no trash to recover from: `DatasetController.delete` goes to `dfso.rm`, and `FileSystem.delete` bypasses trash.

Blast radius is wider than direct user calls, because the SDK itself uses `upload(..., overwrite=True)` internally for agent scripts, requirements, wheels and runners (`python/hsml/model_serving.py:607,616,820,837`).

## The fix

Check the policy **before** the removal, in `_assert_upload_allowed`, using two reads the SDK can already make:

1. `VariableApi._get_upload_policy()` for the configured policy.
2. `UsersApi._get_current_user()` for the caller's cluster roles, from `GET /users/profile`.

The policy alone settles two of the three cases, so the profile is fetched **only** for `admins_only`, the one value whose answer depends on who is asking. That also confines the API key scope requirement to that branch: the profile endpoint accepts the `USER`, `PROJECT` or `FEATURESTORE` scope (`UsersResource.java:230`), so the note in `project_api.py:57` about `USER` being required is narrower than the endpoint has been since FSTORE-1738.

Unrecognised, empty and missing policy values permit the upload, matching the backend's own tolerant parsing, so a typo in the configured value cannot block a client that the backend would accept.

When the policy is `admins_only` and the profile cannot be read, the overwrite is refused rather than attempted. Refusing costs the caller an overwrite; guessing costs them the file.

`User` gains a `roles` field, parsed from the profile's `role` collection exactly as `AdminUser` already parses it. A `User` built from an embedded reference, such as a project member, carries no role information and leaves it empty.

### An earlier version of this used a request probe, and it was wrong

The first push checked permission by issuing the gated chunk-probe `GET`. That looked appealing because it asks the authority instead of duplicating the decision, but `UploadService` calls `configureUploader` **before** it validates the flow parameters. A probe carrying deliberately incomplete parameters therefore returns 400 for every caller the policy permits, and the check only worked because it treated any non-403 as permission. It relied on an error response as its success case and would have put a 400 in the server log for every overwrite upload. Replaced.

## One smaller fix on the same path

A refusal was unreadable. `_upload_chunk` special-cased only `110055` (disk space), so a policy refusal surfaced through `RestAPIError`, whose message is `"Metadata operation error: (url: ...)"` followed by the raw HTTP status, body and headers. It now recognises `110056` and reports a sentence, matching how the disk-space code is already handled a few lines above. This matters for a fresh upload, which the preflight does not cover because there is nothing to remove.

## Testing

Tests folded into the existing `TestDatasetApiUploadChunk` class in `python/tests/core/test_dataset_api.py`, following that file's pattern (inline `mocker.patch`, no new fixtures). They cover: `disabled` blocking everyone without reading the profile at all, `admins_only` permitting an admin and blocking a non-admin, the refusal when the role cannot be read, case and whitespace handling, `enabled` / empty / missing / unrecognised all permitting, the profile's `role` collection parsing into `roles`, and the two orderings that matter:

- `overwrite=True` under a blocking policy: `remove` is **never called** and no upload is attempted.
- `overwrite=True` when permitted: still removes, then uploads.

Plus the refusal message, its non-retry, and both constants.

`ruff check` and `ruff format --check` are clean on all five files.

**The repo test suite could not be run locally.** `uv sync --extra dev --all-groups --project python` fails with `hopsworks-docs[cli]==0.1.0` unsatisfiable, which I reproduced on a clean tree, so it predates this change and is environmental. The new tests therefore rely on CI. To avoid asserting behaviour I had not executed, I exercised the same paths against the real `DatasetApi` in a standalone harness with only the HTTP layer stubbed: 21 checks pass, including both "destination not removed" cases and "an admin can still overwrite under admins_only".

## Not addressed here

`copy` and `move` carry the same delete-then-write shape with `overwrite=True`, but neither goes through the upload endpoint, so the upload policy does not apply to them and this change leaves them alone.

Ticket: https://hopsworks.atlassian.net/browse/FSTORE-2086


python/hopsworks_common/core/variable_api.py
```diff
@@ -158,6 +158,22 @@ def _get_loadbalancer_external_domain(
...
+        One of `enabled`, `admins_only` or `disabled`, as enforced by the backend on the dataset
+        upload endpoint.
+
+        Returns:
+            The configured value, or None when the cluster has no policy configured, in which
+                case the backend's own default applies.
+        """
+        try:
+            return self._get_variable("upload_policy")
+        except RestAPIError:
```
robzor92: **The swallow also catches 403, which puts the data loss back.**

The comment below is true of only one of the two errors that land here. `VariablesService.getVar` throws `ServiceErrorCode.VARIABLE_NOT_FOUND` when the row is absent — safe, since the backend then falls back to `enabled` — but it also throws `SecurityErrorCode.REST_ACCESS_CONTROL` (403) when the row's visibility is `ADMIN` and the caller is not:

```java
if (variable.get().getVisibility() == VariablesVisibility.ADMIN && !sc.isUserInRole("HOPS_ADMIN")) {
  throw new HopsSecurityException(RESTCodes.SecurityErrorCode.REST_ACCESS_CONTROL, ...
```

Both become `None`, i.e. "no restriction", so `remove()` runs and the upload 403s. The destination is gone — for exactly the non-admin the policy was blocking.

The only thing preventing that is hopsworks-helm#2201 writing `visibility` 1, which the SDK cannot verify. The `variables` DDL default is `0`, which is `ADMIN` (first enum constant), and logicalclocks.github.io#632 tells operators they may set this from the Configuration page, where the UI seeds USER but offers "Hops admin" as a radio option, and `ConfigurationItem.tsx` lets it be changed on an existing row afterwards.

It is also the opposite decision from the one taken a few lines into `_assert_upload_allowed`, which refuses when the profile cannot be read on the grounds that "refusing costs the caller an overwrite; guessing costs them the file". That reasoning applies here too. Suggest branching on the status / error code — treat only 404 / `VARIABLE_NOT_FOUND` as absent, and refuse on anything else. Either way the comment needs fixing, and no test covers this branch.

python/hopsworks_common/core/dataset_api.py
```diff
@@ -229,6 +234,8 @@ def upload(
                     raise DatasetException(
                         "overwrite=True not supported on a top-level dataset"
                     )
+                # Before the removal, never after: see _assert_upload_allowed.
+                self._assert_upload_allowed(destination_path)
```
robzor92, line 238: **This fixes one trigger of the delete-first shape, not the shape.**

The harm the PR body describes — "The destination is gone and there is nothing left to restore" — is equally reachable five lines from the code this PR adds to `_upload_chunk`: `DATASET_ERROR_CODE_UPLOAD_DISK_SPACE` at :404 is raised after `remove()` has already run. Same for a mid-upload network failure, chunk-retry exhaustion, and a quota rejection. Overwriting a 10 GB file onto a full cluster loses it before this change and after it.

Write-then-swap closes all of them with primitives already in this file: upload to `destination_path + ".upload-tmp"`, then `remove(destination_path)`, then `move(tmp, destination_path)` (:775). The destructive window shrinks from a multi-GB transfer to two metadata operations, and the policy refusal then arrives before anything is removed *without* the client reproducing the backend's decision at all — no policy read, no profile read, no new `User.roles` field, and no fail-open / fail-closed asymmetry to reason about.

The trade-offs are real (peak storage doubles, the temp needs cleaning up on failure, the directory case is more work), so deferring is defensible — but then the PR body should say so, because as written it reads as though the destructive window is closed.

python/hopsworks_common/core/dataset_api.py
```diff
@@ -398,6 +405,14 @@ def _upload_chunk(
                             "Upload failed: HopsFS storage is full. "
                             "Please contact your administrator to free up disk space."
                         ) from re
+                    if (
+                        re.error_code
+                        == DatasetApi.DATASET_ERROR_CODE_UPLOAD_NOT_ALLOWED
+                    ):
+                        raise DatasetException(
+                            "Upload failed: uploading files is not allowed on this cluster. "
```
robzor92, line 413: This mapping only covers `_upload_chunk`. `upload_feature_group` (:849) calls `_upload_request` directly, so a policy refusal on that path still surfaces as the `RestAPIError` debug dump this branch exists to replace. That method is legacy and deprecated, so leaving it alone is probably right — one sentence in the PR body would make it a decision rather than an oversight.

python/hopsworks_common/core/dataset_api.py
```diff
@@ -418,6 +433,49 @@ def _get_flow_base_params(self, file_name, num_chunks, size, chunk_size):
...
+
+        # Only this policy depends on the caller, so the profile is read only here.
+        try:
+            user = users_api.UsersApi()._get_current_user()
+        except RestAPIError as re:
+            # Refusing costs the caller an overwrite; guessing costs them the file.
+            raise DatasetException(
+                "Uploading files is restricted to cluster administrators on this cluster, and "
+                "your role could not be determined, so "
+                f"{destination_path} was left unchanged. An API key with the USER, PROJECT or "
```
robzor92: This branch also fires on a 500 or an expired token, where naming API key scopes points the caller at the wrong thing. It is wrong for in-cluster callers in every case: jobs and Jupyter authenticate with a JWT (`JobsJWTManager` mints audience `api`), so there is no API key to fix. Consider saying the role could not be read, and appending the scope hint only when the failure is a 403.

python/hopsworks_common/core/users_api.py
```diff
@@ -76,6 +76,24 @@ class UsersApi:
     Use [`hopsworks.get_users_api`][hopsworks.get_users_api] to get an instance of this class.
     """
 
+    def _get_current_user(self) -> User | None:
```
robzor92, line 79: Nothing tests this method. `python/tests/core/test_users_api.py` already has everything it needs: `_patch_client(mocker, ...)`, a `_profile()` builder that already returns `"role": [{"groupName": "HOPS_USER"}]`, and the path-asserting idiom (`assert get_call.args[1] == ["admin", "users", 1]`).

A wrong path or verb here would keep every test in the PR green, because both `_assert_upload_allowed` tests mock this method out — the `["users", "profile"]` path and the empty-body `None` case are unverified. About four lines in a file built for exactly this.

python/tests/core/test_dataset_api.py
```diff
@@ -103,6 +104,155 @@ def test_dataset_error_code_upload_disk_space_constant(self):
...
+        )
+        mock_profile = mocker.patch(
+            "hopsworks_common.core.users_api.UsersApi._get_current_user"
+        )
+
+        api._assert_upload_allowed("Resources")
+
+        mock_profile.assert_not_called()
+
+    def test_user_profile_parses_cluster_roles(self):
```
robzor92: This is a `User.from_response_json` test, and the repo already has a home for those: `python/tests/test_user.py`, backed by `fixtures/user_fixtures.json` through the `backend_fixtures` fixture. Inside `TestDatasetApiUploadChunk` it is hard to find and it is not what the class is about.

While here: `test_assert_upload_allowed_permits_admin_under_admins_only` (:186) asserts nothing — it passes if nothing raises. Asserting that the profile *was* read would make it pin the branch its name claims.

python/hopsworks_common/user.py
```diff
@@ -53,6 +55,10 @@ def from_response_json(cls, json_dict: dict[str, Any] | None) -> User | None:
             if "firstname" in json_decamelized:
                 json_decamelized["first_name"] = json_decamelized.pop("firstname")
                 json_decamelized["last_name"] = json_decamelized.pop("lastname")
+            # The profile sends roles as `role`, a list of group objects.
+            role = json_decamelized.pop("role", None)
+            if role:
+                json_decamelized["roles"] = [group.get("group_name") for group in role]
```
robzor92: This is now the second copy of the same parse (`AdminUser._from_decamelized`, :114). The commit message declines the reviewer's None-filter suggestion precisely *because* the two copies have to agree — a shared helper removes that constraint, and then either parser can be hardened without the other drifting.

robzor92 (commented):
Adversarial review — I went after this PR's load-bearing claims against hopsworks-ee#3224, hopsworks-helm#2201, logicalclocks.github.io#632 and the ee auth/variables code. Details inline; two things that do not anchor to a diff line:

**What holds up.** The policy parsing matches `UploadPolicy.fromString` exactly (trim, case, and `enabled` / empty / missing / unrecognised all permitting). The role source agrees with the enforcement point: `sc.isUserInRole("HOPS_ADMIN")` is built from `userUtilities.getUserRoles` for API keys (`ApiKeyFilter.java:70`) and from the same call in the job JWT's `ROLES` claim (`JobsJWTManager.java:66`), which is the `BbcGroup` collection that `UserProfileDTO` exposes as `role` (`UsersBuilder.buildFull:133`) — client and backend cannot disagree. I expected `@JWTRequired(acceptedTokens = {Audience.API})` on `GET /users/profile` to lock out in-cluster callers, but job tokens are minted with audience `api` and Jupyter's is `Audience.API`, so that is fine. The scope claim is accurate, and the extra round trip adds no new requirement: `GET /variables/{id}` needs `ApiScope.PROJECT`, which every external client already spends on `getProjectInfo` at login (`external.py:113`). Dropping the chunk probe, and dropping the `FLOW_PERMANENT_ERRORS` entry in the last commit, were both right.

**One gap with no line to hang it on.** Under a blocking policy a *fresh* directory upload still litters. `mkdir` (:711) posts to the dataset endpoint, and ee#3224 gates only `UploadService.configureUploader`, so `upload("mydir", "Resources")` on a `disabled` cluster creates the whole remote tree and then 403s on the first chunk, leaving empty directories behind. Probably a backend gap rather than this PR's, but worth a line in the ticket family.

robzor92 (approved):

python/hopsworks_common/core/variable_api.py
manu-sj: Fixed this now an error is thrown for all other cases other than not found

python/hopsworks_common/core/dataset_api.py
manu-sj: I think this would be a larger implementational changes. So we should do it out of this PR changes.

I think for large volumes of data this proposed approach might not be scalable.

python/hopsworks_common/core/dataset_api.py
manu-sj: upload_feature_group is deprecated and has to be removed. SO agreed with the comment no point in updating it

PR #1139, opened 2026-09-03
[HWORKS-3197] Stop sending a service list when creating a project

jimdowling:
The client's fixed service list on `POST /project` (`constants.SERVICES.LIST`) was legacy: the backend decides which services a project gets, and the list is why projects created through the client (`hopsworks.create_project`, the first-project prompt in `hopsworks.login()`, the `hops` CLI) never got Superset. The request no longer carries a list, and the constant and its re-exports (`hopsworks.constants.SERVICES`, `hsml.constants.SERVICES`) are deleted.

**Requires the backend change from the same ticket, logicalclocks/hopsworks-ee#3299.** Until that is deployed, a request without a list yields a project with Kafka alone, so this client must not be released ahead of it.

## Verification

- Unit: `tests/core/test_project_api.py` asserts the POST body carries no `services`; the model and predictor suites that import the shared fixtures still pass; ruff and `check_pep8_public.py` pass.
- **Live on jim-teleport** with logicalclocks/hopsworks-ee#3299 deployed: `hopsworks.create_project(...)` from this branch produced `project_services` rows `AIRFLOW, FEATURESTORE, JOBS, JUPYTER, KAFKA, SERVING, SUPERSET, TRINO`; `main`'s client with its fixed list gets the same set minus `SUPERSET`. Test project deleted afterwards.

## Loadtests

No loadtest creates a project through the SDK's `create_project` path; the suite seeds its projects with an explicit REST call and its own service list (`init_api_key.py`, `helpers/users.py`), so it neither covers nor is affected by this change.

### DIFF
--- a/python/hopsworks_common/constants.py
+++ b/python/hopsworks_common/constants.py
@@ -48,7 +48,16 @@ class GIT:
 
 
 class SERVICES:
-    LIST = ["JOBS", "KAFKA", "JUPYTER", "SERVING", "FEATURESTORE", "AIRFLOW"]
+    LIST = [
+        "JOBS",
+        "KAFKA",
+        "JUPYTER",
+        "SERVING",
+        "FEATURESTORE",
+        "AIRFLOW",
+        "TRINO",
+        "SUPERSET",
+    ]
 
 
 class OPENSEARCH_CONFIG:


jimdowling:
Closed in favour of #1143: this PR's recorded head was an "Update branch" merge commit that a later force-push replaced, so it kept rendering that stale diff (the service list being *added*, not removed) and GitHub would not reopen it. #1143 is the same branch with the intended change: the client stops sending a service list.

PR #1140, opened 2026-09-03
[FSTORE-2082] Delete online rows from the vector database

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```markdown
```


python/hsfs/feature_group.py
```diff
@@ -4954,12 +4954,13 @@ def remove_rows(
                       Combining it with `timeout` `0` waits forever, since neither the entry count nor the timeout can end the wait.
                 - key `internal_kafka` and value `True` or `False` in case you established connectivity from your Python environment to the internal advertised listeners of the Hopsworks Kafka Cluster.
             storage: The storage to delete from, `"offline"` or `"online"`, mirroring [`insert`][hsfs.feature_group.FeatureGroup.insert].
-                Left unset it follows the feature group: both stores when it is online-enabled, offline alone when it is not or when its online data lives in the vector database behind an embedding index.
+                Left unset it follows the feature group: both stores when it is online-enabled, offline alone when it is not.
                 Set it to `"offline"` to delete from the offline table only, or `"online"` to delete from the online store only.
+                On a feature group with an embedding index, the online store is the vector database, and the rows are removed from it by primary key.
```
dhananjay-mk, line 4959: **Priority: Medium**

This changes public `remove_rows` behavior in a patch release. The public API convention requires an `Info: ..., ~=X.Y.Z` admonition for patch-level behavior changes. Please add one identifying the first version that supports deleting embedding-index rows from the online vector store.

python/hsfs/feature_group.py
```diff
@@ -5029,7 +5010,10 @@ def remove_rows(
         if delete_online:
             # Requires an OnlineFS (clusterj-onlinefs) that understands the
             # `operation: delete` header (the release shipping the OnlineFS delete
-            # branch onward). Not runtime-gated: OnlineFS is not reachable from the
+            # branch onward), and on a feature group with an embedding index also one
+            # whose vectordb committer routes tombstones to a vector database delete
+            # rather than indexing them as documents.
+            # Not runtime-gated: OnlineFS is not reachable from the
```
dhananjay-mk: **Priority: High**

This enables online deletion without negotiating whether the deployed OnlineFS supports delete tombstones. The client compatibility check accepts patch-level differences, while OnlineFS before logicalclocks/clusterj-onlinefs#189 handles every row as a document write. In that mixed-version case, this can replace the existing vector document with tombstone data instead of deleting it.

Could we expose a runtime capability through the backend and fail closed here when it is absent? The deployment-order assumption in this comment does not protect separately released clients or backports.

dhananjay-mk:
## Cross-PR review summary

Reviewed logicalclocks/hopsworks-ee#3290, this PR, logicalclocks/clusterj-onlinefs#189, and logicalclocks/loadtest#1012 together.

- **Priority: High:** The client can send vector-delete tombstones to an older, same-minor OnlineFS that interprets them as document writes. Add runtime capability negotiation and fail closed when deletion is unsupported. [Inline discussion](https://github.com/logicalclocks/hopsworks-api/pull/1140#discussion_r3967073466)
- **Priority: Medium:** Document the patch version introducing this public API behavior using the required `Info: ..., ~=X.Y.Z` admonition. [Inline discussion](https://github.com/logicalclocks/hopsworks-api/pull/1140#discussion_r3967073458)
- **Priority: Medium:** Align OnlineFS handling of no-primary-key deletes so the log message, ingestion counters, tests, and RonDB behavior report the same outcome. [Inline discussion](https://github.com/logicalclocks/clusterj-onlinefs/pull/189#discussion_r3967073470)

The backend bulk-delete implementation and the OnlineFS last-operation-wins handling look sound. I found no security issue or exposed secret in the reviewed diffs.

python/hsfs/feature_group.py
bubriks: You're right about the failure mode, and it's worse than a no-op: against an OnlineFS with the RonDB delete branch but without clusterj-onlinefs#189, the tombstone is indexed as a document and overwrites the real vector under that id. I've made the comment say that explicitly instead of leaving it at "no-op-to-corrupting", and say which case the deployment ordering does *not* cover — a client pinned to an SDK newer than the cluster it talks to, which the major.minor check at connection time lets through.

On gating it, though, I'd rather not, for two reasons.

There is nothing to negotiate against today. `variables/versions` exposes zookeeper, grafana, logstash, kibana, ndb, hive2, spark, flink, kafka, opensearch, tensorflow and hopsworks — no `onlinefs`, and the backend PR (hopsworks-ee#3290) adds only the OpenSearch `batchDelete`, no capability variable. So this would mean a new backend variable plus the chef/helm plumbing to populate it from the deployed OnlineFS, in this PR's scope.

And the gate would be in the wrong place if we added it only here. The `operation: delete` header is not new in this PR — it shipped in FSTORE-2068 (#1069), which is already in 5.1.0-RC1, and the RonDB path has exactly the same exposure: an older OnlineFS upserts a row holding only the primary key. Failing closed on the embedding path while the RonDB path stays open would give the narrower of the two problems the protection. If we want capability negotiation for online deletes, I think it deserves its own ticket covering both stores, where the backend variable and the helm/chef side can be done properly rather than bolted onto the vectordb half.

Happy to file that follow-up if you agree — or if you'd rather block this PR on it, say so and I'll do the backend variable here.

python/hsfs/feature_group.py
bubriks: I don't think this one is a patch-level change: `remove_rows` itself is new in the same release line. It was added by FSTORE-2068 (#1069) at `5.1.0.dev1`, and `main` is still `5.1.0.dev1` — so the online delete and this change to it have never been in a released version together with different behavior. There is no version boundary for an `Info: ..., ~=X.Y.Z` admonition to mark.

This is not going to `branch-5.1`, so 5.1.0's docs will describe `remove_rows` refusing an online delete on an embedding feature group and that will stay true for 5.1.0. If we do decide to backport it later, I'll add the admonition with the actual patch version then.

dhananjay-mk (approved):

PR #1155, opened 2026-09-08
[FSTORE-2109] Stop training dataset writes from pinning a HybridRowQueue per shuffle partition

robzor92:
[FSTORE-2109](https://hopsworks.atlassian.net/browse/FSTORE-2109)

## Problem

A `fv_*_create_fv_td_*` job fails on a **60-row** dataset:

```
org.apache.spark.memory.SparkOutOfMemoryError: [UNABLE_TO_ACQUIRE_MEMORY] Unable to acquire 65536 bytes of memory, got 0
  at org.apache.spark.util.collection.unsafe.sort.UnsafeInMemorySorter.<init>
  at org.apache.spark.sql.execution.SortExec.createSorter
```

With `coalesce=True` the training dataset is written by a single task. Any Python UDF sitting **below** the `Coalesce` is then evaluated once per upstream shuffle partition, and each evaluation creates a `HybridRowQueue` that pins a 64 MB page until the task ends. `HybridRowQueue` is a Scala `case class` whose fields are identical for every queue in a task, so `TaskMemoryManager.consumers` (a `HashSet`) registers only the first — the rest are neither attributed nor spillable. ~30 non-empty partitions exhaust a 4 GB executor.

Memory sizing is not a fix: a training dataset with all 200 partitions non-empty would need ~23 GB of executor memory. This is behind the 8 failing `python_driver/test_fully_qualified_names.py` tests and `python_driver/test_snowflake_schema::test_workflow` in the helm nightly.

Two sites reach it, and both are fixed here.

## 1. The time-series split filter

`_time_series_split` filtered every split through the `convert_event_time_to_timestamp` Python UDF. The epoch-millisecond column is now built with native expressions, dispatched on the column type:

| type | UDF today | native |
| --- | --- | --- |
| timestamp | `int(dt.timestamp() * 1000)` | `unix_millis(col)` |
| **timestamp_ntz** (what Spark 4 actually yields) | same | `timestampdiff(MILLISECOND, TIMESTAMP_NTZ '1970-01-01 00:00:00', col)` |
| date | UTC midnight in ms | `datediff(col, '1970-01-01') * 86400000` |
| int / bigint | `0 → null`, ≤10 digits → `× 1000`, else as-is | `when`/`length` equivalent |

Any other type keeps the UDF, and so does an event time whose case does not match the column: the type is resolved by scanning the schema, since `StructType.__getitem__` is case-sensitive where `col()` follows `spark.sql.caseSensitive`.

**timestamp_ntz is the case that matters.** Spark 4 reads a feature group's event time as `timestamp_ntz`; `TimestampNTZType` is not a subclass of `TimestampType` and `unix_millis` rejects it outright, so a timestamp/date/bigint dispatch silently falls back to the UDF and nothing changes — verified on the cluster, where the first revision of this branch still OOMed with `BatchEvalPython` in the plan. `FeaturestoreConstants.EVENT_TIME_FEATURE_TYPES` describes the *feature* type, not the Spark type the query read yields. `test_time_series_split_timestamp_ntz` pins this: it asserts the executed plan has no `BatchEvalPython` node, and fails if the branch is removed.

Both operands of the `timestampdiff` are NTZ, so the result is zone-free by construction and independent of `spark.sql.session.timeZone`. The alternatives either depend on the session zone or work by accident:

| form | session tz UTC | session tz America/New_York |
| --- | --- | --- |
| `unix_millis(col)` | AnalysisException | AnalysisException |
| `unix_millis(col.cast("timestamp"))` | match | off by the zone offset |
| `unix_millis(to_utc_timestamp(col, "UTC"))` | match | off by the zone offset |
| `unix_millis(to_timestamp_ltz(col, "UTC"))` | match | match, but the `"UTC"` is a *format pattern* that Spark ignores for non-string input; it relies on `ToTimestamp` passing NTZ micros through on a compatibility path |
| `timestampdiff(MILLISECOND, TIMESTAMP_NTZ '1970-01-01', col)` | match | match |

(`F.timestamp_diff` exists only on pyspark >= 4, so the expression goes through `F.expr`, which also works on 3.5.)

## 2. The unsplit path

An unsplit training dataset ran `_fit_and_transform` first and `coalesce(1)` afterwards, so a **transformation function's** UDF sat below the merge — python and pandas mode alike, since `ArrowEvalPython` shares the evaluator factory.

The merge cannot simply move ahead of the fit: approximate statistics depend on the partitioning they are computed over. On 1M rows, profiling the same data over 200 partitions versus one gives percentiles 249/499/749 vs 250/500/750 and `std_dev` differing in its last bits, so fitting on a coalesced frame would refit every transformation on different statistics *and* change the persisted training-dataset statistics. `repartition(1)` is not order-preserving, and materializing the transformed frame does not scale.

So the engine takes the reshaping as `pre_transform` and applies it once the statistics are fitted and before the transformations run. For a chained fit the statistics stages are planned up front (`_plan_statistics_stages`), so the hook lands after the last stage that profiles anything: no profiling pass moves and every transformation still executes exactly once. Earlier stages sit behind their statistics barrier, which is already persisted, so the write task reads them from cache instead of re-evaluating their UDFs. **No `cache`, `persist` or `checkpoint` is added.**

## Verification

**On the cluster** (hopsworks 5.1.0-SNAPSHOT, Spark 4.1.3.0), re-running the exact `create_fv_td` job that fails in the nightly with the SDK swapped in the driver, same config and sizing (driver 2048m / executor 4096m):

| run | plan | outcome |
| --- | --- | --- |
| unmodified SDK | `BatchEvalPython [convert_event_time_to_timestamp]` | `UNABLE_TO_ACQUIRE_MEMORY` |
| control (wrapper, patch skipped) | same | `UNABLE_TO_ACQUIRE_MEMORY` |
| timestamp/date/bigint only | still `BatchEvalPython`, no `unix_millis` | `UNABLE_TO_ACQUIRE_MEMORY` |
| **with timestamp_ntz** | `unix_millis` ×7, **0** `BatchEvalPython` | **SUCCEEDED**, 167 s |

On the successful run the executors show zero `HybridRowQueue` acquisitions and zero `PythonUDFWithNamedArgumentsRunner` invocations, against 64 MB queues and ~2.16 GB of untracked memory on every failing run.

**Output equality.** Row-for-row identical for timestamp_ntz, date from 1970 on, bigint, int-seconds, `0`/NULL and negative epochs, on pyspark 3.5.5 **and** 4.1.1, under session zones UTC **and** America/New_York. `0` was already treated as missing rather than an error, because the UDF's `if not event_time: return None` shadowed its own `ValueError`. Two deliberate differences, both places where the UDF was wrong: for `timestamp` the UDF's value followed the Python **worker's OS zone** (`TimestampType.fromInternal` is `datetime.fromtimestamp`), not the session zone, so it was off by the zone offset on a non-UTC worker; and dates before 1970 came out 1 ms late (`date.timetuple()[:7]` hands `tm_wday` to `datetime()` as microseconds, so `date(1969, 12, 31)` gave -86399999 instead of -86400000). The native form is UTC in both cases, which is what the cluster's executors run anyway. The unsplit change reproduces the pre-change row-and-value hash (`751561226b8ec847`) on 1M rows with the same fitted statistics.

**Performance**, local A/B on the same data (the fix also removes the UDF's per-row cost):

| rows | split filter, parallel write | split filter, `coalesce(1)` |
| --- | --- | --- |
| 100k | 4.1 s → 2.7 s | 5.4 s → 3.1 s |
| 1M | 7.2 s → 3.1 s | 12.0 s → 4.2 s |
| 10M | 37.6 s → 5.0 s (**7.5×**) | 82.8 s → 13.9 s (**6.0×**) |

Unsplit transform path, 1M rows: 20.1 s → 6.9 s at default pages, and `UNABLE_TO_ACQUIRE_MEMORY` → writes all 1M rows on a 3 GB pool with 64 MB pages.

It does **not** buy predicate pushdown — measured `PushedFilters: []` for both the UDF and the native form at every size, since the comparison is on a function of the column.

**Tests.** Eight new: an NTZ event time produces a plan without `BatchEvalPython`; an event-time name whose case differs from the column still reaches the UDF fallback instead of raising `KEY_NOT_EXISTS`; statistics are profiled on all four partitions while only the transformed frame is merged; the plan keeps the Python UDF above the `Coalesce` (and below it without the hook, with identical rows); call order is `["fit", "pre", "apply"]`, `["pre", "apply"]` with no statistics, and `["fit", "apply", "fit", "pre", "apply"]` chained with exactly one application per stage; and a split dict is rejected rather than silently mis-shaped. The four pre-existing `coalesce.call_count` tests pass unmodified. Suites: `test_spark.py` 266 passed (only the 5 pre-existing avro/kafka failures), transformation engine 153 passed, python engine 234 passed.

## Not covered here

The split path fits its statistics on a coalesced frame (it merges before `_fit_and_transform`) while the unsplit path fits on the full partitioning, so the same data through the two paths yields different approximate percentiles. Pre-existing, and changing it would change existing split-TD outputs — worth its own ticket.

No upstream JIRA covers the `HybridRowQueue` case-class/HashSet de-duplication; the Spark-side fix is identity equality for the queue, which would fix every path including user code, and should be reported separately.

[FSTORE-2109]: https://hopsworks.atlassian.net/browse/FSTORE-2109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/hsfs/engine/spark.py
+++ b/python/hsfs/engine/spark.py
@@ -57,13 +57,18 @@ try:
         col,
         concat,
         current_timestamp,
+        datediff,
+        expr,
         from_json,
+        length,
         lit,
         monotonically_increasing_id,
         regexp_replace,
         row_number,
         struct,
         udf,
+        unix_millis,
+        when,
     )
     from pyspark.sql.types import (
         ArrayType,
@@ -80,6 +85,7 @@ try:
         StringType,
         StructField,
         StructType,
+        TimestampNTZType,
         TimestampType,
     )
 
@@ -1165,15 +1171,20 @@ class Engine:
             # Statistics are always refit (training_dataset_version not
             # passed): an unsplit in-memory training dataset retrieved by
             # version is not consistent.
+            # coalesce(1) must land between the fit and the transformations
+            # (FSTORE-2109), so the engine applies it.
             dataset = transformation_function_engine.TransformationFunctionEngine._fit_and_transform(
                 training_dataset,
                 feature_view_obj,
                 dataset,
                 transformation_context=transformation_context,
+                pre_transform=(
+                    (lambda frame: frame.coalesce(1))
+                    if training_dataset.coalesce
+                    else None
+                ),
             )
 
-            if training_dataset.coalesce:
-                dataset = dataset.coalesce(1)
             path = training_dataset.location + "/" + training_dataset.name
             return self._write_training_dataset_single(
                 dataset,
@@ -1353,13 +1364,14 @@ class Engine:
                 "Given event time should be in `datetime`, `date`, `str` or `int` type"
             )
 
-        # registering the UDF
-        _convert_event_time_to_timestamp = udf(
-            convert_event_time_to_timestamp, LongType()
-        )
+        # A Python UDF below the coalesced write pins a 64 MB HybridRowQueue
+        # per shuffle partition (FSTORE-2109); use a native expression when
+        # the column type allows one.
+        ts_col = self._event_time_epoch_millis(dataset, event_time)
+        if ts_col is None:
+            ts_col = udf(convert_event_time_to_timestamp, LongType())(col(event_time))
 
         result_dfs = {}
-        ts_col = _convert_event_time_to_timestamp(col(event_time))
         for split in training_dataset.splits:
             result_df = dataset.filter(ts_col >= split.start_time).filter(
                 ts_col < split.end_time
@@ -1368,6 +1380,40 @@ class Engine:
                 result_df = result_df.drop(event_time)
             result_dfs[split.name] = result_df
         return result_dfs
+
+    @staticmethod
+    def _event_time_epoch_millis(dataset, event_time):
+        """Epoch milliseconds of the event time column as a native expression.
+
+        Returns None when the column type has no native form.
+        """
+        # not schema[event_time]: that subscript is case sensitive, col() is not
+        event_time_type = next(
+            (f.dataType for f in dataset.schema.fields if f.name == event_time), None
+        )
+        if isinstance(event_time_type, TimestampNTZType):
+            # unix_millis rejects timestamp_ntz; with both operands NTZ the
+            # result does not depend on spark.sql.session.timeZone.
+            quoted = event_time.replace("`", "``")
+            return expr(
+                "timestampdiff(MILLISECOND, TIMESTAMP_NTZ '1970-01-01 00:00:00', "
+                f"`{quoted}`)"
+            )
+        if isinstance(event_time_type, TimestampType):
+            return unix_millis(col(event_time))
+        if isinstance(event_time_type, DateType):
+            # midnight UTC of that day, independent of the session zone
+            days = datediff(col(event_time), lit("1970-01-01")).cast(LongType())
+            return days * lit(86_400_000)
+        if isinstance(event_time_type, (IntegerType, LongType)):
+            # up to ten digits is seconds
+            value = col(event_time).cast(LongType())
+            return (
+                when(value == 0, lit(None).cast(LongType()))
+                .when(length(value.cast("string")) <= 10, value * lit(1000))
+                .otherwise(value)
+            )
+        return None
 
     def _write_training_dataset_splits(
         self,
--- a/python/hsfs/core/transformation_function_engine.py
+++ b/python/hsfs/core/transformation_function_engine.py
@@ -1248,6 +1248,7 @@ class TransformationFunctionEngine:
         transformation_context: dict[str, Any] | None = None,
         n_processes: int | None = None,
         training_dataset_version: int | None = None,
+        pre_transform: Callable[[Any], Any] | None = None,
     ) -> (
         dict[str, pd.DataFrame | pl.DataFrame | TypeVar("pyspark.sql.DataFrame")]
         | pd.DataFrame
@@ -1268,10 +1269,19 @@ class TransformationFunctionEngine:
             transformation_context: Context variables passed to the transformation functions.
             n_processes: Number of worker processes for applying the transformations.
             training_dataset_version: When given, statistics are retrieved from the backend for this version instead of being fit.
+            pre_transform: Applied to a single frame after the statistics are fit and before the transformations run, so a `coalesce(1)` neither changes the profiled statistics nor evaluates the transformations once per parent partition (FSTORE-2109). Not supported for split dictionaries.
 
         Returns:
             The transformed dataset, in the same shape as `dataset`.
         """
+        # no caller passes the hook for splits; without this it would be
+        # silently dropped
+        if pre_transform is not None and isinstance(dataset, dict):
+            raise ValueError(
+                "pre_transform is only supported when `dataset` is a single "
+                "dataframe, not a dictionary of splits."
+            )
+
         if training_dataset_version is not None:
             # The statistics of an existing training dataset version already
             # exist in the backend: bind them, then every frame (train
@@ -1280,6 +1290,8 @@ class TransformationFunctionEngine:
                 training_dataset, feature_view_obj, training_dataset_version
             )
             if not isinstance(dataset, dict):
+                if pre_transform is not None:
+                    dataset = pre_transform(dataset)
                 return TransformationFunctionEngine._transform(
                     feature_view_obj, dataset, transformation_context, n_processes
                 )
@@ -1297,6 +1309,7 @@ class TransformationFunctionEngine:
                 dataset,
                 transformation_context=transformation_context,
                 n_processes=n_processes,
+                pre_transform=pre_transform,
             )
 
         # The train split is fit on and transformed first: fitting binds the
@@ -1361,6 +1374,7 @@ class TransformationFunctionEngine:
         | TypeVar("pyspark.sql.DataFrame"),
         transformation_context: dict[str, Any] | None = None,
         n_processes: int | None = None,
+        pre_transform: Callable[[Any], Any] | None = None,
     ) -> pd.DataFrame | pl.DataFrame | TypeVar("pyspark.sql.DataFrame"):
         """Fit the required statistics on the train data and return it transformed.
 
@@ -1385,6 +1399,7 @@ class TransformationFunctionEngine:
             feature_dataframe: The train data.
             transformation_context: Context variables passed to the transformation functions.
             n_processes: Number of worker processes for applying the transformations.
+            pre_transform: Applied to the train data once the statistics are fit and before the transformations run.
 
         Returns:
             The transformed train data.
@@ -1424,6 +1439,7 @@ class TransformationFunctionEngine:
                 label_encoder_features,
                 transformation_context,
                 n_processes,
+                pre_transform=pre_transform,
             )
 
         if statistics_features:
@@ -1439,6 +1455,10 @@ class TransformationFunctionEngine:
             for tf in feature_view_obj.transformation_functions:
                 tf.transformation_statistics = stats.feature_descriptive_statistics
 
+        # statistics are fit on the frame as read; reshape only now
+        if pre_transform is not None:
+            feature_dataframe = pre_transform(feature_dataframe)
+
         return TransformationFunctionEngine._transform(
             feature_view_obj, feature_dataframe, transformation_context, n_processes
         )
@@ -1453,6 +1473,7 @@ class TransformationFunctionEngine:
         label_encoder_features: set[str],
         transformation_context: dict[str, Any] | None,
         n_processes: int | None,
+        pre_transform: Callable[[Any], Any] | None = None,
     ) -> pd.DataFrame | pl.DataFrame | TypeVar("pyspark.sql.DataFrame"):
         """Fit statistics over a chained DAG and return the transformed data.
 
@@ -1471,6 +1492,7 @@ class TransformationFunctionEngine:
             label_encoder_features: The statistics features consumed by encoders.
             transformation_context: Context variables passed to the transformation functions.
             n_processes: Number of worker processes for applying each stage.
+            pre_transform: Applied once the last stage that fits statistics has been profiled and before its transformations run.
 
         Returns:
             The fully transformed train data.
@@ -1485,20 +1507,20 @@ class TransformationFunctionEngine:
         )
 
         provisional_statistics: list = []
-        provisioned_features: set[str] = set()
         working_dataframe = feature_dataframe
         original_columns = list(feature_dataframe.columns)
         statistics_engine = training_dataset._statistics_engine
         previous_barrier = None
 
-        remaining = list(graph.nodes)  # topological order
-        while remaining:
-            stage, remaining = TransformationFunctionEngine._next_statistics_stage(
-                remaining, provisioned_features
-            )
-            features_to_fit = {
-                f for tf in stage for f in tf.hopsworks_udf.statistics_features
-            } - provisioned_features
+        stages = TransformationFunctionEngine._plan_statistics_stages(
+            list(graph.nodes)  # topological order
+        )
+        # pre_transform goes after the last stage that fits anything
+        pre_transform_stage = max(
+            (i for i, (_, features_to_fit) in enumerate(stages) if features_to_fit),
+            default=0,
+        )
+        for stage_index, (stage, features_to_fit) in enumerate(stages):
             if features_to_fit:
                 # Present in working_dataframe: raw features from the start,
                 # intermediate features materialized by earlier stages. Profiled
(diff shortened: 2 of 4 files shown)


robzor92:
Validated on a dev cluster (hopsworks 5.1.0-SNAPSHOT, Spark 4.1.3.0) by re-running the exact `create_fv_td` job that fails in the nightly, and the first revision of this PR **did not fix it** — pushed a second commit that does.

**What the first revision missed:** on Spark 4 the event-time column reads back as `timestamp_ntz`, and `TimestampNTZType` is not a subclass of `TimestampType`, so the dispatch fell through to the UDF fallback — the plan still had `BatchEvalPython [convert_event_time_to_timestamp]` and the executors still pinned `HybridRowQueue` pages. `unix_millis` also rejects that type outright, so the fallback was the only thing keeping the job from an `AnalysisException`.

`FeaturestoreConstants.EVENT_TIME_FEATURE_TYPES` (`TIMESTAMP`, `DATE`, `BIGINT`) describes the *feature* type, not the Spark type the query read yields — that is what misled the original analysis.

**The added branch:** `unix_millis(to_timestamp_ltz(col, lit("UTC")))`. A naive timestamp is a UTC wall clock, which is how the UDF treated it; pinning the zone keeps it independent of `spark.sql.session.timeZone`. The two alternatives both match the UDF *only* when the session zone happens to be UTC:

| form | session tz UTC | session tz America/New_York |
| --- | --- | --- |
| `unix_millis(col)` | AnalysisException | AnalysisException |
| `unix_millis(col.cast("timestamp"))` | match | **off by the zone offset** |
| `unix_millis(to_utc_timestamp(col, "UTC"))` | match | **off by the zone offset** |
| `unix_millis(to_timestamp_ltz(col, "UTC"))` | match | match |

**Cluster evidence**, same job config (`fv_pk_1fg_query_1_create_fv_td_...046`, driver 2048m / executor 4096m), SDK swapped in the driver only:

| run | plan | outcome |
| --- | --- | --- |
| unmodified SDK (loadtest) | `BatchEvalPython [convert_event_time_to_timestamp]` | `UNABLE_TO_ACQUIRE_MEMORY` |
| control, wrapper but no patch | same | `UNABLE_TO_ACQUIRE_MEMORY` |
| revision 1 (timestamp/date/bigint only) | still `BatchEvalPython`, no `unix_millis` | `UNABLE_TO_ACQUIRE_MEMORY` |
| revision 2 (+ `timestamp_ntz`) | `unix_millis` ×7, **0** `BatchEvalPython` | **SUCCEEDED**, 167 s |

On the successful run the executors show zero `HybridRowQueue` acquisitions and zero `PythonUDFWithNamedArgumentsRunner` invocations, against ~2.16 GB of untracked memory and 64 MB queues on every failing run — which does confirm the queue mechanism in the issue, just not the type coverage.

Row-for-row equivalence re-run with a `timestamp_ntz` case added: identical rows and order for timestamp, timestamp_ntz, date, bigint, int-seconds, `0`/NULL and negative epochs, on pyspark 3.5.5 **and** 4.1.1, under session tz UTC **and** America/New_York.

python/hsfs/engine/spark.py
```diff
@@ -1353,13 +1370,42 @@ def convert_event_time_to_timestamp(event_time):
...
+        # 64 MB HybridRowQueue page until the task ends. The queues are not tracked
+        # by Spark's TaskMemoryManager (HybridRowQueue is a case class, so they all
+        # compare equal), so ~30 non-empty partitions exhaust the executor and the
+        # write fails with UNABLE_TO_ACQUIRE_MEMORY on 60 rows (FSTORE-2109).
+        event_time_type = dataset.schema[event_time].dataType
+        if isinstance(event_time_type, TimestampNTZType):
+            # Spark 4 reads Hudi/Delta event times as timestamp_ntz, and unix_millis
+            # rejects that type. A naive timestamp is a UTC wall clock, like the UDF
+            # treated it, so pin the zone instead of leaning on the session's.
+            ts_col = unix_millis(to_timestamp_ltz(col(event_time), lit("UTC")))
```
robzor92: The second argument of `to_timestamp_ltz` is a *format pattern*, not a zone. For a `timestamp_ntz` input it is ignored entirely: `lit('Europe/Stockholm')`, `lit('not-a-zone-or-format')` and `lit('yyyy')` all produce identical results, and the same call on a string column throws `INVALID_DATETIME_PATTERN: Unrecognized datetime pattern: 'UTC'`. The correct result comes from `ToTimestamp` returning the NTZ micros unchanged for non-string input, a compatibility path Spark says it should ideally not accept. So the comment above describes a mechanism that does not exist, and the expression is one Spark change away from returning null per row (non-ANSI) and writing empty splits.

Zone-free by construction, matches the UDF under both session zones tested:

```python
# pyspark >= 4.0
ts_col = timestamp_diff("MILLISECOND", lit(datetime(1970, 1, 1)).cast("timestamp_ntz"), col(event_time))
# pyspark 3.5
ts_col = expr(f"timestampdiff(MILLISECOND, TIMESTAMP_NTZ '1970-01-01 00:00:00', `{event_time}`)")
```

If the current form stays, the comment should say what it actually relies on.

python/hsfs/engine/spark.py
```diff
@@ -1353,13 +1370,42 @@ def convert_event_time_to_timestamp(event_time):
...
+        # The event time is compared as epoch milliseconds. Do it with native
+        # expressions for the column types Hopsworks allows as event time
+        # (timestamp, date, bigint): a Python UDF here turns the per-split filter
+        # into a BatchEvalPythonExec, and with `coalesce=True` the coalesced write
+        # task then runs it once per upstream shuffle partition, each time pinning a
+        # 64 MB HybridRowQueue page until the task ends. The queues are not tracked
+        # by Spark's TaskMemoryManager (HybridRowQueue is a case class, so they all
+        # compare equal), so ~30 non-empty partitions exhaust the executor and the
+        # write fails with UNABLE_TO_ACQUIRE_MEMORY on 60 rows (FSTORE-2109).
+        event_time_type = dataset.schema[event_time].dataType
```
robzor92: `StructType.__getitem__` is case-sensitive while `col()` honours `spark.sql.caseSensitive=false`, so `event_time="EVENT_TIME"` on an `event_time` column used to work and now raises `PySparkKeyError [KEY_NOT_EXISTS]` before any job runs. Hopsworks lowercases feature names, so mostly theoretical, but `next((f.dataType for f in dataset.schema if f.name == event_time), None)` falling into the existing UDF `else` branch keeps the old behaviour for free.

python/hsfs/engine/spark.py
```diff
@@ -57,13 +57,18 @@
...
         current_timestamp,
+        datediff,
         from_json,
+        length,
         lit,
         monotonically_increasing_id,
         regexp_replace,
         row_number,
         struct,
+        to_timestamp_ltz,
```
robzor92: `to_timestamp_ltz`, `unix_millis` and `TimestampNTZType` are all pyspark >= 3.5. Fine for 5.0 and 5.1, but this `try` block ends in `except ImportError: pass`, so on an older runtime every pyspark name silently goes undefined and the engine fails later with NameErrors. Only relevant if a 4.x backport is planned.

robzor92 (commented):
Adversarial review: tried to break both halves on pyspark 4.1.1 (UTC and America/New_York session zones, ANSI on and off, and a non-UTC worker OS zone), plus mutation-testing the new tests. The `pre_transform` half holds; the split half has one expression that works for the wrong reason and one untested branch, and two claims in the description do not hold.

**Needs a change**

1. **The `"UTC"` in `to_timestamp_ltz(col, lit("UTC"))` is a format pattern, not a zone, and is ignored for a `timestamp_ntz` input.** `'Europe/Stockholm'`, `'not-a-zone-or-format'` and `'yyyy'` all give byte-identical results; the same call on a string column throws `INVALID_DATETIME_PATTERN: Unrecognized datetime pattern: 'UTC'`; without the second argument the result becomes session-zone dependent. It works because `ToTimestamp` returns the NTZ micros unchanged for non-string input, a compatibility path Spark's own source says it should ideally not accept. If Spark ever applies the pattern, non-ANSI mode returns null for every row and the splits are silently empty. `timestampdiff(MILLISECOND, TIMESTAMP_NTZ '1970-01-01 00:00:00', col)` matches the UDF under both session zones and is zone-free by construction (both operands NTZ): `F.timestamp_diff` on pyspark 4, `F.expr` on 3.5. Inline comment on the line.

2. **Nothing pins the NTZ dispatch.** Deleting the `TimestampNTZType` branch, which reproduces the revision-1 no-op described above, leaves all 9 existing `_time_series_split` / `_split_df` tests green. One more case next to `test_time_series_split_timestamp` that casts `event_time` to `TimestampNTZType()` and asserts `"BatchEvalPython" not in result[split]._jdf.queryExecution().executedPlan().toString()` fails on that mutation.

**Claims in the description that do not hold** (the native form is the correct one in both cases, so no code change)

3. Not row-for-row identical for dates before 1970: the UDF does `datetime(*event_time.timetuple()[:7])`, which passes `tm_wday` as microseconds. Positive epochs truncate it away, negative ones round toward zero, so `date(1969-12-31)` gives -86399999 from the UDF and -86400000 natively.

4. For `TimestampType` the UDF's value depends on the Python worker's OS zone (`TimestampType.fromInternal` is `datetime.fromtimestamp`), not the session zone: with the OS at Europe/Stockholm every row is off by 1-2 h. The session-zone table tests a knob the UDF never read, and one the engine pins to UTC in `Engine.__init__` anyway. Executors on the cluster are UTC, so nothing changes there. Worth stating as "the UDF was wrong off-UTC and the native form fixes it".

**Low**

5. `dataset.schema[event_time]` is case-sensitive while `col()` is not; a name-case mismatch that used to work now raises `PySparkKeyError [KEY_NOT_EXISTS]` before any job runs. Inline comment.

6. The new imports raise the Spark engine's floor to pyspark 3.5; the module-level `except ImportError: pass` turns that into NameErrors later on an older runtime. Only matters for a 4.x backport.

**What held up**

- int/bigint matches the UDF for 0, negatives, 10 vs 11 digits and `IntegerType`, ANSI on and off; date matches for every date from 1970 on.
- All three `pre_transform` mutations fail the intended tests: merge before the fit, merge at stage 0 in the chained path, merge after the last fitting stage's transformations. The stage replay is pure.
- No join, repartition or second UDF sits in the Spark transform path, so the single partition survives to the write.

python/hsfs/engine/spark.py
robzor92: Replaced with `expr("timestampdiff(MILLISECOND, TIMESTAMP_NTZ '1970-01-01 00:00:00', `col`)")` (both operands NTZ, so zone-free by construction; `F.expr` because `F.timestamp_diff` only exists on pyspark >= 4). The comment now says why `to_timestamp_ltz` is *not* used. Backticks in the column name are escaped. Still matches the UDF row-for-row on 3.5.5 and 4.1.1 under UTC and America/New_York.

python/hsfs/engine/spark.py
robzor92: Done: `next((f.dataType for f in dataset.schema.fields if f.name == event_time), None)`, so a case mismatch falls into the UDF branch as before. Added `test_time_series_split_event_time_case_mismatch`, which fails against the subscript form with `KEY_NOT_EXISTS`.

python/hsfs/engine/spark.py
robzor92: Leaving as is: 5.0 and 5.1 both ship pyspark >= 3.5 and no 4.x backport is planned for this. `to_timestamp_ltz` is gone from the imports anyway; `unix_millis` and `TimestampNTZType` remain, both 3.5+.

robzor92:
Addressed in the amended commit:

1. `to_timestamp_ltz(col, lit("UTC"))` → `timestampdiff(MILLISECOND, TIMESTAMP_NTZ '1970-01-01 00:00:00', col)` via `F.expr` (works on 3.5 and 4.x). The code comment explains why the ltz form is avoided.
2. Added `test_time_series_split_timestamp_ntz`: casts the event time to `TimestampNTZType` and asserts neither `BatchEvalPython` nor `ArrowEvalPython` appears in the executed plan. Disabling the NTZ branch fails exactly that test; the other 10 split tests stay green, as you found.
3. / 4. Description and commit message now state both as places where the UDF was wrong (worker OS zone for `timestamp`, 1 ms late before 1970 for `date`) and the native form fixes them, rather than claiming row-for-row identity there.
5. Type lookup is a schema scan instead of `schema[event_time]`; `test_time_series_split_event_time_case_mismatch` pins the UDF fallback for a name-case mismatch.
6. Left as is; pyspark 3.5 is the floor on both shipping branches.

`test_spark.py`: 266 passed, only the 5 pre-existing avro/kafka failures.

manu-sj (approved):

PR #1167, opened 2026-09-13
[FSTORE-2116] Batch inference and training data anchored on a caller-supplied spine_df

jimdowling:
https://hopsworks.atlassian.net/browse/FSTORE-2116

A feature view is anchored on its root feature group, so
`get_batch_data(start_time, end_time)` only returns rows for times that feature
group has already observed. Inference over the future has none: a view joining
air quality observations to weather observations and forecasts returns nothing
for tomorrow, even though the weather feature group holds the forecast. The
workaround is to read the forecast feature group directly, which bypasses the
view's joins, feature selection and transformations.

`get_batch_data(spine_df)` takes a dataframe of entities carrying a prediction
time per row instead, treats it as an inference spine, and anchors the query on
that spine. Every other feature group is looked
up ASOF each prediction time. Offline feature groups only; online serving is
unchanged.

## What this repo contributes

The public API is `spine_df` on `get_batch_data` and on the six training-data
methods, plus explicit `primary_key` / `event_time` overrides on the batch read.
`PredictionTimes` accepts a cron expression, a named interval with an offset, or
an explicit list of timestamps, and expands to a bounded horizon of times; its
`cross()` builds the frame. The cron parser follows the Vixie dialect and is
implemented in the SDK, so no dependency is added.

Serving keys cross with those times to give one row per entity and prediction time,
each carrying a row id that fixes the identity and order of the output frame.
The Python engine stages the spine as Parquet under the caller's `Resources`
directory and deletes it after the read. Spark registers the rows as a temporary
view and writes no file, which is why the wire form advertises a file only when
one was staged.

Keys and the prediction time are included by default for this call, where they are
excluded by default for an ordinary batch read, because without them the frame does
not say which row belongs to which entity or day. Passing `False` leaves them out. A backend that does not echo the spine back is
refused before execution rather than silently answered with historical rows.

## Verification

Built and cluster-verified on `jim-asof-inference` (Proxmox) on 2026-09-13.

The acceptance criterion is the air quality example in
[`featurestorebook/mlfs-book`](https://github.com/featurestorebook/mlfs-book):
its direct read of the weather forecast feature group replaced by a feature
view call, returning the four weather columns the model consumes cell-for-cell
identical to the direct read, 7 rows, one per date, in prediction-time order.
That holds on the cluster, along with 13 of 13 checks in the acceptance suite:
required serving keys derived, root feature group pruned out of all 16 spine
queries, unknown entity and unknown-city cases, a partially filled frame, feature-age
bounds, row order under shuffled and duplicated rows, spine file lifecycle,
and no caller value in the flyingduck log. Spark was exercised separately as an
in-cluster PySpark job, 7 of 7 checks, returning the same column list as the
query service path.

| Suite | Result |
| --- | --- |
| hopsworks-ee query package, incl. `TestSpineSqlGenerator` | 296 pass, 0 fail |
| hopsworks-api unit tests | 4938 pass, 13 skipped; see the third-round note on how to run them |
| flyingduck | 532 pass; 5 pre-existing ibis import failures |
| hopsworks-front | 33 pass, lint clean |
| hopsworks-helm `.github/test_lint.sh` | clean |

### Loadtests

Run `516eb33f` against `jim-asof-inference` with the out-of-cluster runner, the
modified SDK driving the modified backend, so the combination under test is the
one that would ship. 15 files across both the python and pyspark drivers,
selected to cover every path this change touches: `test_batch`,
`test_batch_extra_filter`, `test_batch_event_time_excluded`,
`test_pit_join_filter_consistency`, `test_pit_lookback_window`,
`test_pit_join_pushdown`, `test_training_data` and `test_feature_group_read`.

**49 passed, 1 failed, 1 skipped, 1h 25m.**

The failure, `test_batch_event_time_excluded`, is pre-existing and unrelated. It
is proven so: the EAR was rebuilt from the same tree with this change removed
and redeployed, and the test failed identically, same error, same five
occurrences. While that baseline was deployed it also served as a positive
control for the client-side capability gate, which refused the spine query
rather than answering it with historical rows.

Not covered by that run: the `e2e_scale` marker, the data source tests
(`CONNECTORS` was not set), and every workflow outside the feature store.

### Known gaps

- The conformance suite runs on the query service. Its Spark copy is committed in
  logicalclocks/loadtest#1038 and has not run, because the in-cluster Python
  environment carries the released client. Spark itself was exercised separately as
  an in-cluster job with the modified client, 7 of 7 checks.
- The capacity benchmark covers spine rows and row width. History depth, key skew,
  multi-child views and the Spark engine are unmeasured.
- Nested joins and `RIGHT`/`FULL` joins are refused rather than supported.

## Second review round

Two more defects, neither caught by any test.

**`max_feature_age` silently applied no bound when its key named no feature group.**
The client checked only that each value was a `timedelta`; the backend looked the name
up in a map with a wildcard fallback and returned no bound on a miss. A typo therefore
disabled the guard and returned rows carried forward for ever, with nothing to notice.
Both ends refused an unknown name from then on, the backend with a new code 332. The
per-name form has since been replaced by a single bound for the whole view, which removes
the class of mistake entirely: there is no name left to get wrong. Code 332 is withdrawn
with it, and the client now refuses a zero or negative bound instead, the only remaining
way to ask for a bound that can match nothing.

**The change record contradicted the code.** The earlier commit and body said keys and
event time default to excluded for a spine read. The code includes them by default,
which is why the notebook passes `False`. The specification had it right, so the error
was confined to the change record, where a reviewer would have taken it as the
contract. Corrected in the follow-up commit and in this body.

Both gaps that made those defects possible are also closed. The nine-shape DuckDB
binding harness is committed, as a golden file in `hopsworks-ee` and as the test that
binds it in `flyingduck`. The result-level conformance suite is committed to
logicalclocks/loadtest#1038 and passes on the query service.

Writing that suite found two wrong expectations of mine before it found anything else,
which is the argument for asserting on frames rather than on SQL. A root feature group
whose observations stop does not return NULL for a future prediction time, it carries
its last observation forward. And an inclusive one-day bound still reaches yesterday's
row, so it does not null a one-day gap.

### Capacity benchmark

Run on `jim-asof-inference`, query service, one child feature group, 1000 entities,
30 days of history. RSS is summed over both query service pods, sampled every 3 s.

| Spine rows | Columns | Passed value | Latency | flyingduck RSS |
| --- | --- | --- | --- | --- |
| 1 000 | 2 | none | 2.1 - 2.8 s | 608 MiB flat |
| 10 000 | 2 | none | 2.2 s | 608 MiB flat |
| 100 000 | 2 | none | 3.5 s | 608 MiB flat |
| 1 000 000 | 2 | none | 7.2 - 7.6 s | 608 to 612 MiB |
| 10 000 | 103 | 1 KiB per row | 6.7 s | 821 MiB flat |

The provisional limits are kept, now measured. A spine at `max_rows` answers in about
eight seconds and moves the query service's resident set by 4 MiB, so the row cap is
not where the risk is. Row width is: 10 000 rows of 103 columns cost more than 100 000
narrow rows, which is what the byte and column caps are for.

### Loadtest regression, current build

Run `bfe7550e` against `jim-asof-inference` with the out-of-cluster runner, the modified
client driving the backend and query service built from this branch and its siblings. 17
files across both drivers: `test_batch`, `test_batch_extra_filter`,
`test_batch_event_time_excluded`, `test_pit_join_filter_consistency`,
`test_pit_lookback_window`, `test_pit_join_pushdown`, `test_training_data`,
`test_feature_group_read`, and the new conformance suite.

**50 passed, 2 failed, 1 skipped, 1 h 23 m.**

Both failures are understood and neither is a regression.

- `test_batch_event_time_excluded` fails with `Values list "fg0" does not have a column
  named "trans_freq"`, the same error and the same pre-existing failure proven earlier by
  rebuilding the EAR with this change removed and redeploying.
- `test_asof_batch_inference` on the pyspark driver **passes**. The job runs the client
  baked into the Spark image, which is still the released one, so the run points it at a
  wheel built from this branch (`LOADTEST_CLIENT_WHEEL`): the driver puts the wheel first
  on `sys.path`, where zipimport shadows the installed client without rebuilding the image.
  Unset, the case stays xfail. The test asserts the line the suite prints on reaching its
  end and that the driver picked the wheel up, so a job that exits 0 having skipped the
  suite cannot read as a pass.

Not covered by this run: the `e2e_scale` marker, the data source tests (`CONNECTORS` was
not set), and every workflow outside the feature store.

## Also here: `entry` deprecated in favour of `serving_keys`

Filed as FSTORE-2118 and folded into this branch rather than shipped separately, so
the API lands consistent in one release instead of disagreeing with itself for one.
That ticket is closed as duplicated.

`get_feature_vector`, `get_feature_vectors`, `get_inference_helper` and
`get_inference_helpers` now accept `serving_keys`. All four take the same dictionary,
so leaving the inference helpers behind would have replaced one inconsistency with
another.

```python
fv.get_feature_vector(serving_keys={"id": 1})   # preferred
fv.get_feature_vector(entry={"id": 1})          # works, warns
fv.get_feature_vector({"id": 1})                # unchanged
```

`serving_keys` takes the position `entry` held and `entry` moves to the end of the
signature, so positional callers are unaffected. Passing `entry` emits a
`DeprecationWarning` naming the replacement. Passing both is refused: they are one
argument, so two values are a caller bug rather than a preference to resolve quietly.

The resolver is a private static method rather than a module-level function, which is
not a style preference. A module-level definition in `feature_view.py` displaces the
`FeatureView` class on its generated API reference page, which breaks every
cross-reference to a `FeatureView` member and fails the documentation build in strict
mode. That is how it was found.

Out of scope and left alone: `TrainingDataset.get_serving_vector` and
`get_serving_vectors`, the legacy serving path superseded by feature views, and the
internal vector server plumbing, which is not public.

## Also here: training data from the same rows

`training_data`, `train_test_split`, `train_validation_test_split` and the three
`create_*` methods now take `spine_df` as well. The observation that prompted it:
anchoring a query on a caller-supplied dataframe is what a spine group does, and the
inference spine is the late-bound form of that, so a feature view no longer has to have
been created with a spine group to build training data from a labels frame.

```python
train_x, test_x, train_y, test_y = feature_view.train_test_split(
    test_size=0.2,
    spine_df=labels,   # keys, an event time per row, and the label
)
```

Two differences from a batch read. A training row is one entity at one moment rather
than an entity scored repeatedly, so a schedule is rarely what builds the frame. And
columns the view does not define are carried through to
the output untouched, which is how the label rides along; a batch read stays strict,
because inference has no labels and an unrecognised column there is worth catching.

`spine_df` and `spine` both replace the left side of the query, so passing both is
refused. An event-time window or a lookback is suppressed under a spine, because the
spine is the population and a window on top of it would drop rows the caller asked for.

A materialisation keeps its staged spine file, because its Spark job reads it after the
call returns. A sweeper reclaims those, and the ones a read failed to delete. Before
this there was no sweeper, so the warning the client logged on a failed delete described
one that did not exist.

### Two defects the cluster found

Adding the training case to the conformance suite surfaced two bugs, neither reachable
from the cases that existed before.

- **Spine-bound join keys were missing from the physical read.** They come from the
  spine's bindings rather than from a `Join`, so the walk that collects needed features
  never saw them, and DuckDB reported `Values list does not have a column named <key>`.
  It stayed hidden while the root feature group was always pruned. A training read keeps
  the root, and the same shape was reachable from `get_batch_data` with a kept root whose
  keys are not selected, so this was a latent bug in the batch path too.
- **The Parquet file and its declared schema disagreed for passthrough columns.** The two
  were computed separately and a passthrough column fell back to the view's schema, which
  does not define it, so the file was written as a string. One function decides both now.

## Naming

The dataframe parameter is `spine_df`, on `get_batch_data` and on all six training-data
methods. The frame is a spine: it replaces the left side of the query and may carry passed
features and labels as well as keys, so naming it for the keys described only part of it.

The methods whose argument is a dict of keys keep `serving_keys`, where that name is exact:
`get_feature_vector`, `get_feature_vectors` and the two inference helper methods. The feature
view's own `serving_keys` property and the feature logging argument of the same name are
unrelated and unchanged.

## API consolidation

`spine_df` is the only spine-related argument of `get_batch_data`. Two parameters came
off it, both because the API was saying the same thing in more than one way.

**`prediction_times` is gone, losslessly.** The times were always expressible as a column
of the frame, and the training-data methods already required them there, so the batch path
was the odd one out. `PredictionTimes` remains as the thing that builds a schedule from a
cron expression, a named interval or a list, and it gains `cross()`. That matters: the row
order is a documented contract, entities as given and ascending in time within each, and
leaving callers to hand-roll the cross product would have made that contract theirs.

```python
schedule = PredictionTimes.every("daily", offset="00:00", start=tomorrow, count=7)
df = fv.get_batch_data(spine_df=schedule.cross(entities, event_time="date"))
```

**`max_feature_age` moved onto the feature view.** It could not follow the times into the
frame, because it bounds how stale a lookup may be, which is a property of the lookup and
not of a row. Deleting it would have removed the only defence against an as-of join
carrying a stale value forward silently. As a property of the view it applies to every read
anchored on a spine, batch inference and training data alike.

```python
fv = fs.create_feature_view(name="air_quality_fv", query=query,
                            max_feature_age=datetime.timedelta(days=1))
```

It is written once at creation and is read-only afterwards, because an object-only bound is
lost the moment the view is fetched again: a training run and the inference pipeline that
scores against it are separate processes, and each would have had to remember to set the
same value. That is the training/serving skew the bound exists to prevent.
`fv.max_feature_age` reads it back as a `timedelta`, or `None` when the view is unbounded.

It is one value for the whole view rather than a bound per feature group. A per-group map
needs a `VARCHAR` column holding JSON, a name to validate against the view's feature groups,
and an error code for the name that matches nothing; a single bound needs an `INT` of
seconds and no validation beyond positive, and no pipeline asked for the extra resolution.
The column is `feature_view.max_feature_age_secs INT NULL`, added by migration V100 and
nullable, so an existing view keeps the unbounded behaviour it has today. The spine carries
`maxFeatureAgeSecs`, and neither the query service nor the chart needed a change.

The conformance suite proves the bound reaches training data, not just batch: unbounded,
the gap day is filled by carry-forward; read through a view created with the bound, exactly
that day comes back NULL and the caller's label is untouched.

It also proves the bound survives a round trip, which the rest of the suite cannot. Every
other bounded case reads through the object that created the view, and that object carries
the bound in memory, so all of them would pass unchanged if the column were never written.
The round-trip case fetches each bounded view again, checks it is a different object,
checks the bound came back, and reads through the fetched view to confirm the value that
came back is the one the read is held to.

Verified on `jim-asof-inference` after applying V100 and redeploying: the suite passes, and
the rows behind it read 43200 and 86400 seconds against the two bounded views with NULL
against the unbounded one.

### `spine_df` accepts a Spark DataFrame

Under the Spark engine, `get_batch_data` and all six training-data methods now take a Spark
DataFrame as well as pandas, polars or a list of dicts. It is collected to the driver with
`toPandas()` rather than kept distributed, which is the route the Spark path already took:
the rows are handed to `createDataFrame` to register the session temporary view, so a pandas
spine travelled the same way. A spine is one row per entity per prediction time and is capped
at a million rows, so it is bounded by construction.

The Python engine refuses a Spark frame and names the fix, because it has no session to
evaluate one. `PredictionTimes.cross` refuses one as well: it builds a pandas frame, and
pandas answers a Spark frame with `DataFrame constructor not properly called`, which tells
the caller nothing about what to do instead.

One dependency is worth recording. `toPandas()` returns timestamps tz-naive in the Spark
session timezone, and the spine reads them back as UTC. Those agree only because the Spark
engine pins the session timezone to UTC when it starts. A test pins that assumption, so
unpinning it fails here rather than shifting every event time silently.

### One definition of what counts as a Spark DataFrame

The Spark checks above first went through a predicate I added to `hopsworks_common/util.py`,
matching on class name plus module prefix. That was the fourth such definition in the tree,
and the weakest of them. `hopsworks_common/spark_connect_utils` already holds the canonical
one, an `isinstance` against both the classic `pyspark.sql.DataFrame` and the Spark Connect
`pyspark.sql.connect.dataframe.DataFrame`, and `engine/spark.py` delegates to it with a
comment saying it exists so that a single rule decides what counts as a Spark DataFrame
across the codebase.

The one I added is deleted. `read_primary_keys`, the spine conversion and
`PredictionTimes.cross` all call the shared predicate. Name matching would have got Spark
Connect wrong as soon as the class moved, and a fourth rule is a fourth thing to keep in step
with the other three.

### The latest feature values, without maintaining a list of entities

`fv.get_root_fg()` returns the feature group the view's query is anchored on, and
`fg.read_primary_keys()` returns that feature group's distinct primary key values, one row per
entity.

```python
fg = feature_view.get_root_fg()
now = datetime.datetime.now(datetime.timezone.utc)

spine_df = PredictionTimes.of([now]).cross(fg.read_primary_keys(), event_time=fg.event_time)
latest = feature_view.get_batch_data(spine_df=spine_df)
```

`read_primary_keys` returns entities and no time, so it is not a spine on its own and passing
it straight to a read is refused. Crossing it with one instant is what makes it one, and
capturing that instant once is what has every entity read as of the same moment. There is no
implicit "as of now" anywhere in this API, because a call that defaulted to the wall clock
would return different rows on a re-run and a training dataset materialized from it could
never be reproduced.

It reads the key columns of the whole feature group and takes the distinct rows, so the cost
scales with the feature group rather than with the number of entities, and it covers the
root's keys only: a joined feature group keyed on something the root does not carry comes back
NULL.

It deliberately does not call `self.read()`. That path applies the scheduler's
`HOPS_START_TIME` / `HOPS_END_TIME` window, so a caller asking for every entity inside a job
would get whatever interval that job happens to be processing, with nothing in the result to
say so.

`dataframe_type` behaves as it does on `read`, within each engine's own limits. `"default"` is
a Spark DataFrame under the Spark engine and a pandas one under the Python engine, `"pandas"`
works on both, and `"polars"` is a Python-engine type that the Spark engine's converter
rejects exactly as it does for `read`. `"numpy"` and `"python"` are refused here although
`read` returns them, because an ndarray and a list of lists have no notion of a distinct row.

### Spine groups are deprecated

`spine_df` supersedes them. A spine group has to be chosen when the feature view is created
and cannot be added afterwards, so a view built without one can never be driven by a caller's
rows, and a view built with one requires `spine=` on every read. `get_or_create_spine_group`
is deprecated, and passing `spine` warns on all eight methods that accept it. There is no
in-place conversion: migrating a view means recreating it from a query with no spine group on
the left. The `hops-fg` and `hops-batch-inference` skills drop them.

The `SpineGroup` class carries a deprecation block in its docstring but is not decorated. A
class-level `@deprecated` in `hopsworks_apigen` wraps every member `inspect.getmembers`
returns, inherited ones included, which rewrites an inherited `@staticmethod` into an instance
method and breaks it at the first call. Reproduced in isolation, and worth fixing in that
package rather than working around twice here.

### Two defects the Spark run found

Both were reachable only by running the conformance suite under Spark, which had never
happened: its driver was xfail from inception, so every assertion in the suite was written
pandas first.

- `read_primary_keys` dispatched a Spark DataFrame to the pandas branch and failed with
  `DataFrame.drop_duplicates() got an unexpected keyword argument 'ignore_index'`. A Spark
  frame answers to `drop_duplicates` as well as to `distinct`, so a capability check picks the
  wrong one. Dispatch is by type now, Spark first.
- `PredictionTimes.cross` failed inside pandas on a Spark frame instead of refusing it.

A third defect was found by unit tests rather than by the cluster, and would have been worse.
Making `max_feature_age` read-only left it in the attribute list `update_from_response_json`
assigns through, and creating a feature view posts the object and applies the response through
that same method, so it would have raised `AttributeError` on `create_feature_view` itself
rather than only on a refresh.

### Verification

hopsworks-api: **4938 passed, 13 skipped**, run from the repo root with `PYSPARK_SUBMIT_ARGS`
naming the Spark avro package. Run from `python/`, or without that variable, thirteen tests
fail on relative fixture paths and a missing avro jar, neither of which is related to this
change.

On `jim-asof-inference` the conformance suite passes on both engines against this client:
python driver 1 passed in 61 s, pyspark driver 1 passed in 281 s. The pyspark run drives the
wheel built from this branch through `LOADTEST_CLIENT_WHEEL`, because the Spark image still
carries the released client.

## Review round: thirteen findings, applied

A review of this change against itself. Two findings were worth blocking on.

### A column name reached SQL unvalidated

`InferenceSpineResolver` held the spine's table name, its Parquet basename and every declared
column *type* to a closed grammar, and said so in a comment: no request string reaches SQL. The
column **names** were not checked. A passthrough column is the one kind with no feature in the
store behind it, so its name arrived exactly as the caller wrote it, and
`SpineSqlGenerator.addPassthrough` renders it into a backtick identifier in both the DuckDB and
the Spark rendering. A name like ``label`, (SELECT token FROM other.secrets) AS `x`` therefore
reached the statement that runs on the signed path.

What stopped it was flyingduck: `_load_spine` refuses a column name that is not an identifier
before it creates the temp table. That is real defence in depth, and it is also sanitisation
living in the service that trusts the backend rather than in the one that builds the statement.
Every declared column name is now matched against `[A-Za-z_][A-Za-z0-9_]*` in `validateShape`,
with a test for an injected passthrough name and one for an injected ordinary name.

### The staleness bound was taken from the request

`max_feature_age` is stored on the feature view and read-only precisely so that a training set
and the inference reads scored against it cannot be held to different limits. The backend was
reading it from `InferenceSpineDTO.maxFeatureAgeSecs`, which is to say from the request. Nothing
read `feature_view.max_feature_age_secs`. The spine arrives on the generic query-constructor
endpoint, which has no feature view in scope, so the backend could not have checked it without
new plumbing. The guarantee held for callers who used this SDK and for nobody else.

The bound is no longer on the wire. The spine names the feature view it anchors, by featurestore
id, name and version, and the resolver reads the bound from that view's own row. The featurestore
is resolved with `getFeaturestoreForProjectWithId(project, true, id)`, so naming a view the
caller cannot reach is refused rather than answered, and a spine that names no view is refused
with the new code 332. The property now holds however the request was built.

### The rest

- **The sweeper walked every project every hour.** An `exists` and a `listStatus` per project,
  on a cluster where almost none has ever staged a spine: thousands of NameNode round trips an
  hour to find nothing. It globs `/Projects/*/Resources/.hopsworks_spine/*` once instead;
  `DistributedFileSystemOps` gains a `globStatus` for it.
- **`get_root_fg` was annotated `FeatureGroup`** while its docstring promised an
  `ExternalFeatureGroup` or a `SpineGroup` for views built on those. Both are siblings under
  `FeatureGroupBase`, so the two documented cases raised `TypeCheckError` under
  `HOPSWORKS_RUN_WITH_TYPECHECK`, which the CI unit-test job sets. The tests only covered a plain
  feature group, which is why nothing caught it.
- **The `max_feature_age` signature still advertised the withdrawn per-feature-group map.** Both
  dict arms always raised. The signature is what the API reference renders, so the published docs
  would have documented a form that cannot work. It now reads `timedelta | int | None`, and a
  dict is refused with a message naming the replacement.
- **`recreate_training_dataset` told callers to use `spine_df` and did not accept it.** A
  training dataset built from a spine was silently rebuilt anchored on the root feature group,
  under the same version, because the frame is not recorded with the dataset. It takes `spine_df`
  now, and the docstring says what happens if you leave it out.
- **The ceilings were enforced only after the upload.** The row and column limits are checked
  before the frame is typed, written and staged, and `PredictionTimes.cross` checks the product
  of entities and times before building the frame at all. They mirror the shipped defaults; the
  backend stays authoritative.
- **`_PASSTHROUGH_TYPE` missed the pandas nullable dtypes.** `Int64`, `Float64` and `boolean` are
  what a label column with missing values is; they fell through to `"string"` and then failed on
  the Arrow cast.
- Smaller: the bindable-column check no longer counts passthrough columns, so a frame of an event
  time and a label is refused by the check meant to catch it; the misspelled
  `SplineDataFrameTypes` becomes `DeprecatedSpineTypes`, since one character from
  `SpineDataFrameTypes` is too close to read; a duplicated polars class leaves an `isinstance`
  tuple; `PASSTHROUGH_TYPES` becomes the set it always was.

### Two CI jobs were already failing

Not found by the review, which read local test runs rather than the PR's checks. `ruff format`
wanted two files, and the PEP-8 / `@public` check refused `InferenceSpine.arrow_table` and
`write_parquet`: public-named methods that are internal plumbing. They are `_arrow_table` and
`_write_parquet` now, and both jobs pass locally the way CI runs them.

### Known limitation, stated rather than left implicit

The spine limits are per request. There is no cap on how many spine reads one user or project may
have in flight, so on a shared cluster a single caller can occupy the query service with repeated
large reads. Admission control is recorded on the ticket as ENT-PIT-5 and is not in this change;
the user guide now says so where it lists the limits.

### Verification after the round

Client suite **4946 passed, 13 skipped**, run from the repo root. Five more of this branch's
tests were failing in the optional-dependency matrix jobs, which are gated behind lint and so
had been reporting as skipped: four Spark spine tests that build a DataFrame from rows, which
pyspark refuses below pandas 2.2, and one polars test importing polars directly. Both now skip
when the prerequisite is absent. Both previously-red CI jobs are green: `ruff format`,
and the PEP-8 / `@public` check. A third, `docsig`, had never run because the format step ahead
of it failed first; it had eleven findings this branch introduced, ten docstrings listing a
parameter in a position its signature does not have it in, and one missing `Returns`. Fixed and
clean.

On `jim-asof-inference`, against a backend rebuilt from hopsworks-ee#3339, the conformance suite
passes on both drivers: **2 passed, 5m35s**.

## External review round: eleven findings, applied

An external review (Codex) of the finished change found eleven defects, six of them release
blockers, and four wider gaps. All eleven are fixed on the branches; none was answered in a
comment. The two that would have shipped wrong data:

- **Materialised training dropped `spine_df` at two boundaries.** `_create_training_dataset`
  accepted the frame and did not pass it on, and the Spark job that materialises a split or an
  external-sink dataset rebuilds its query from the feature view, so the spine could not reach
  it at all. Every `create_*` call with `spine_df` built the ordinary historical population while
  batch inference used the caller's rows: a training/inference population mismatch, silent. The
  spine now travels in the job configuration, validated against the view's query by the same
  resolver the read path uses (file existence and READ permission included), and the job reads
  the staged Parquet back and anchors on it. A file that is gone fails the job rather than
  building another population. Training datasets record `spine_anchored` (V104), so a version
  built from a spine cannot later be read or recreated from the view's own rows.
- **The file check did not establish READ.** `exists` and `getFileStatus` succeed with traverse
  permission on the directories alone, and the signed query-service read then ran as the
  superuser. The resolver now calls Hadoop's `access(path, READ)` as the requester before
  signing.

The other nine: a root filter naming a joined feature rendered inside the root's subquery and
failed to bind on both engines (predicates are now routed to the feature group they reference;
one spanning two groups is refused); event times of different types compared raw, and an integer
prediction time was parsed as nanoseconds (the spine is written in the root's event time type,
integers are epoch milliseconds, and a lookup of a different type is converted with the engine's
own function); tie-break primary keys and the event time were missing from the physical
projection; an explicit lookback was silently dropped with a spine (it now composes as a candidate
filter); Spark frames were collected before the row ceiling applied; the query service
materialised the whole spine file before checking its row count (footer first, load bounded);
Spark inferred the spine schema and refused an all-NULL key (it now registers the typed Arrow
table the Python path writes); a passthrough column could shadow a joined feature's output name;
and a dropped lookup's array or struct feature came back as an untyped NULL (each engine's own
type spelling is rendered).

The full list with dispositions is in the workspace's `FSTORE-2116_review.md`.

### Verification after the round

Cluster `jim-asof-inference`. Backend rebuilt and redeployed from the branch, with **V104** applied
and recorded in `flyway_schema_history`; the query service rolled to a rebuilt image carrying the
loader change. Conformance suite, both drivers, with a wheel built from the branch:
**pyspark driver 1 passed** (the branch wheel on the job's `sys.path`), **python driver 1 passed**
(4m43s). The python driver's job case needs the branch entrypoint and client inside the Spark
image, which still carries the released ones, so for the run `fs_py_job_util` was pointed at a
HopsFS copy of `hsfs_utils.py` that puts the branch wheel first on `sys.path`. A scripted
reproduction of that case showed the job configuration carrying `spine`, the job running the branch
client, a six-row frame twelve hours off every root row producing 3 + 3 rows, and the dataset
recorded as spine-anchored.

Suites: hopsworks-ee query package and Arrow Flight controller tests pass, checkstyle clean;
hopsworks-api 4985 passed, 13 skipped, plus the fifteen WINDOWED golden shapes bound against a local Spark
4.1.1; flyingduck 32 spine tests and the fifteen ASOF shapes bound against DuckDB 1.5.2; docs
strict build, markdownlint and snakeoil clean.

Not covered: the job path on a cluster whose Spark image predates this change now fails at
argument parsing instead of building the wrong population, but that refusal itself was not run on
a cluster (it needs an image with the old entrypoint and a backend with the new op).

### In this repo

`_create_training_dataset` forwards `spine_df` and records `spine_anchored` on the dataset;
`_check_spine_matches` refuses reading or recreating a spine-anchored version without `spine_df`
and recreating a historical one on it; `_compute_training_dataset` refuses a spine-anchored
version without its spine, which also covers the job. `hsfs_utils.create_fv_td` reads the spine
the backend put in the job configuration (`read_staged_spine`: the staged Parquet, in spine order)
and passes it as `spine_df`. `InferenceSpine` writes the event time in the root's own type, reads
integers as epoch milliseconds, refuses other root event-time types, applies the row ceiling
before a Spark collect (`limit(MAX_SPINE_ROWS + 1)`) and before a polars conversion, and refuses a
frame column named like a joined feature. `_get_batch_query` keeps the lookback. The Spark engine
registers the spine view from a pandas frame with a schema declared from the same types the
Parquet file is written with, so an all-NULL key column is accepted; the Arrow-table route was
tried first and failed in the cluster image, whose pyarrow has no timezone database.
`TrainingDataset` carries `spine_anchored` both ways on the wire.

Tests: event-time typing (bigint, date, integer input, unsupported type), shadowing a joined
feature, the polars and Spark ceilings, the typed Spark registration with an all-NULL key,
forwarding and recording on create, lookback retention, the three population checks, the job
entrypoint (`tests/utils/test_hsfs_utils_spine.py`), the training dataset flag, and
`test_spine_sql_shapes_spark.py`, which binds every WINDOWED golden shape against a local Spark.

## Review round from manu-sj, applied

Fifteen inline comments across the eight pull requests, all of them valid and all fixed; each is
answered on its own thread. Two were security-relevant and one was a silent wrong answer:

- **A join prefix reached SQL unvalidated.** It becomes part of an emitted column alias, the alias
  is rendered inside backticks, and the backend rewrites every backtick to a double quote before
  signing, so a prefix carrying either character closed the identifier and the rest of it ran under
  elevated privileges. Feature view and training dataset creation validated it; the query
  constructor, which every batch read goes through, did not. This predates the spine work and the
  point-in-time path renders the same alias, so the check now sits where both paths pass.
- **`max_feature_age` was persisted without validation.** A zero or negative value from a REST
  caller or an older client returned every feature of every spine-anchored read as NULL, with
  nothing to say why. The backend refuses it now, with the wording the client already used.
- **An in-memory training dataset did not record its population.** Only the materialised path set
  `spine_anchored`, so `training_data(spine_df=...)` recorded its version as not spine-anchored and
  reading that version back recomputed it on the feature view's own rows under the same number. The
  three readers take `spine_df` now, so such a version can be read back rather than only refused.

The rest: the client's recognised columns now match the set the backend builds from the stored view;
a keyword the API does not take is refused rather than swallowed, which is how the withdrawn
`prediction_times` reached callers as a confusing error; `PredictionTimes` keeps the precision it is
given and reads an integer as epoch milliseconds, as `spine_df` does; the sweeper deletes only files
this feature wrote and has unit tests; the feature view card renders a call the client accepts and
names serving keys the way the client does; the conformance suite asserts the ticket's acceptance
criterion and stops starting statistics jobs; the user guide lists the variables the chart actually
seeds, one sentence per line.

Client: 4986 passed, 13 skipped from the repo root with `PYSPARK_SUBMIT_ARGS` set; ruff check and format clean.

Head after the round: `179b08ae1`.

## Second external review round: four findings, applied

A second external pass (Codex) found four defects, two of them release blockers. All four are
backend renderings, fixed in logicalclocks/hopsworks-ee#3339: a looked-up feature lost its declared
default in the spine SQL while the online statement kept it; spine key bindings were rebuilt from
the join conditions rather than taken from the feature view's serving keys, which answers for the
wrong entity when a child carries a required key of its own; a typed NULL for a nested type produced invalid DuckDB; and string
literals rendered the same for every engine although Spark reads a backslash as an escape and
DuckDB does not.

What lands here is the coverage that holds them. The conformance corpus grows to sixteen shapes per
engine, two of them new: a looked-up feature that declares a default, and the same feature under a
feature-age bound. The Spark binder executes the first of those against a real session and asserts
that a matched row holding NULL takes the default while a row that matched nothing stays NULL,
which is the distinction the guard exists to preserve. The query service asserts the same rows
through DuckDB, so the two engines are pinned to one another rather than each to its own
expectation.

Both migrations are renumbered again, to **V103** and **V104**. V100 through V102 were taken
while this branch was open, by HWORKS-3242, HWORKS-3240 and HWORKS-2895, and a duplicate version
merges without a conflict and only fails at the next install. The content is unchanged.

The full list with dispositions is in the workspace's `FSTORE-2116_review.md`.

### Verification after the round

hopsworks-api **4996 passed, 13 skipped**, plus **18 shapes** bound against a local Spark 4.1.1
(sixteen text comparisons and two executing). flyingduck 108 passed, including the same sixteen
shapes against DuckDB 1.5.2. Backend query package and checkstyle clean.

Cluster `jim-asof-inference`. Backend rebuilt and redeployed from the branch; the query service
rolled to a rebuilt image. The conformance suite gains two cases for the round and passes on both
drivers: **python driver 1 passed** (5m54s), **pyspark driver 1 passed** (5m24s), the Spark job
running the branch client from a wheel.

The first case appends a feature that declares a default to a feature group whose rows were
written before it, so the rows a lookup matches hold no value. A matched row reads as the default,
a prediction before the first row reads as NULL, and under a feature-age bound a row whose only
candidate is stale reads as NULL rather than taking the default.

The second names a child's key from the spine under the prefix the view publishes, with a
different value in the root's own column of that name, so a lookup bound to the root's column
would answer for the other entity. Writing it found that the over-keyed shape the review described
cannot be created at all: `QueryController` refuses a join covering less than the whole of the
child's primary key, so the reachable shape is a join from a non-key root column. Two resolver unit
tests were written around the refused shape and now describe the reachable one.

Head after the round: `7cc97bf00`. Conformance cases in logicalclocks/loadtest#1038 at `5a57963`.

## Third external review round: four findings, applied, and a fifth found on the cluster

A third external pass (Codex) found four defects, two of them release blockers. All are backend
changes, in logicalclocks/hopsworks-ee#3339: serving keys from two occurrences of one feature group
were combined, which was a regression from the round before; feature-vector prepared statements
bypassed the corrected literal dialect, so a STRING default outside ASCII failed the statement in
MySQL; an empty-string default was treated as no default; and a LIKE pattern carried no escape
clause, which the engines resolve differently.

Running the last of those on a cluster found a fifth. `makePitQueryAsof` chose its dialect from the
query's engine flag, and despite its name that flag is true for the Python engine, which is exactly
the engine that reads through the query service, so the statement was always rendered with the
backslash-doubling renderer meant for Spark and Hive. The per-dialect literals added the round
before never reached the path they were added for. Neither the unit tests nor the previous round's
probes could see it, because they all name a dialect themselves.

### Verification after the round

Cluster `jim-asof-inference`, both drivers, with the two cases this round added: **python driver
1 passed** (5m57s), **pyspark driver 1 passed** (5m36s). Backend query package 278 passed,
checkstyle clean; hopsworks-api 4998 passed, 13 skipped, plus 20 Spark-bound shapes; flyingduck's
20 spine shape tests, three executed in DuckDB 1.5.2.

Head after the round: `4fe6014ab`.

## Related pull requests

Merge in this order. The chart is the only writer of `hopsworks.variables`, so
the backend reads its enum defaults until the chart lands.

1. logicalclocks/hopsworks-ee#3339 and logicalclocks/hopsworks-helm#2343, together
2. logicalclocks/flyingduck#248
3. logicalclocks/hopsworks-api#1167
4. logicalclocks/hopsworks-front#2103, independent of logicalclocks/hopsworks-front#2106
5. logicalclocks/loadtest#1038 and logicalclocks/logicalclocks.github.io#653
6. featurestorebook/mlfs-book#56, the air quality example that consumes it

https://claude.ai/code/session_01F22onm6M6KvU686d5imQw5


python/hsfs/core/feature_view_engine.py
```diff
@@ -588,6 +652,9 @@ def _get_training_data(
             # picks it up. The lookback rides on the persisted training dataset
             # and comes back with `td_updated` regardless of whether we just
             # created it or fetched an existing version.
+            query_spine = self._training_spine(feature_view_obj, spine_df, spine)
+            if training_dataset_version:
+                self._check_spine_matches(td_updated, query_spine)
```
manu-sj, line 662: `_get_training_data` creates the metadata at line 600 without setting `training_dataset_obj.spine_anchored`, unlike `_create_training_dataset` at line 517. Every in-memory `training_data`, `train_test_split` or `train_validation_test_split` call with `spine_df` therefore records its version as not spine-anchored. `get_training_data(training_dataset_version=N)` and the two split readers then pass this check and recompute in memory anchored on the root feature group: a different population under the same version, silently. Setting the flag before `_create_training_data_metadata` fixes the record; a unit test like `test_create_training_dataset_forwards_the_spine_and_records_it` for this path would pin it. One decision follows: `get_training_data`, `get_train_test_split` and `get_train_validation_test_split` have no `spine_df` parameter, so once the flag is recorded an in-memory spine-anchored version can only be refused on read. Either the three readers take `spine_df`, or the guide should say such versions are recomputed through `training_data(spine_df=...)`. The loadtest case never re-reads by version, which is why the cluster run passed.

python/hsfs/constructor/prediction_times.py
```diff
@@ -0,0 +1,522 @@
...
+
+    Build one with [`cron`][hsfs.constructor.prediction_times.PredictionTimes.cron],
+    [`every`][hsfs.constructor.prediction_times.PredictionTimes.every], or
+    [`of`][hsfs.constructor.prediction_times.PredictionTimes.of].
+
+    Example:
+        ```python
+        fv.get_batch_data(
+            spine_df=pd.DataFrame([{"country": "SE", "city": "Stockholm", "street": "Sveavagen"}]),
+            prediction_times=PredictionTimes.every("daily", offset="08:00", count=7),
```
manu-sj: This example still passes `prediction_times=` to `get_batch_data`, which this PR removed, and it renders into the API reference. `_from_user_input` (line 363) has no caller left either; only `TestFromUserInput` keeps it alive.

python/hsfs/constructor/prediction_times.py
```diff
@@ -0,0 +1,522 @@
...
+        pass
+    elif isinstance(value, date):
+        value = datetime(value.year, value.month, value.day)
+    else:
+        raise TypeError(
+            f"Expected a datetime, date, ISO-8601 string or epoch seconds; got {type(value)!r}."
+        )
+    if value.tzinfo is not None:
+        value = value.astimezone(tz)
+    return value.replace(tzinfo=None, microsecond=0, second=0)
```
manu-sj: `_to_local_naive` drops seconds and microseconds from every input, including explicit `PredictionTimes.of()` instants. The documented latest-values recipe `PredictionTimes.of([now])` therefore reads as of the start of the current minute and misses rows written in the last 59 seconds, while the guide says only sub-millisecond precision is dropped. Line 118 also reads an integer as epoch seconds, where `spine_df` reads integers as epoch milliseconds. Keeping sub-minute precision for `of()` and aligning (or refusing) integer input would make the two consistent.

python/hsfs/constructor/inference_spine.py
```diff
@@ -0,0 +1,469 @@
...
+            raise FeatureStoreException(
+                f"Feature group `{root_fg.name}` keeps its event time `{self._event_time}` as"
+                f" `{self._event_time_type}`, which a prediction time cannot be compared with."
+                f" Supported event time types: {sorted(_EVENT_TIME_TYPES)}."
+            )
+        required_keys = {
+            sk.required_serving_key for sk in feature_view.serving_keys if sk.required
+        }
+        root_features = {f.name for f in root_fg.features}
+        recognized = required_keys | root_features | {self._event_time}
```
manu-sj: `root_features` admits every column of the root feature group; the backend (`InferenceSpineResolver.java:172`) accepts only the root's selected features plus keys and event time. A frame with an unselected root column passes here and fails there with a misleading accepted-columns list. See the matching comment on hopsworks-ee#3339.

python/hsfs/feature_view.py
```diff
@@ -1345,6 +1462,7 @@ def get_batch_data(
         extra_filter: filter.Filter | filter.Logic | None = None,
         lookback: FeatureGroupLookback | Lookback | dict[str, Any] | None = None,
         n_processes: int | None = None,
+        spine_df: SpineDataFrameTypes | None = None,
         **kwargs,
```
manu-sj, line 1490: `**kwargs` swallows unknown keyword arguments silently. `prediction_times=` existed on earlier revisions of this branch and is what the hopsworks-front#2103 snippet emits; a caller passing it gets the unrelated 'must carry the prediction time' error. Refusing `prediction_times` (or anything beyond `primary_keys`) with a pointer to `PredictionTimes.cross` would give a direct message.

manu-sj (commented):
Review against the FSTORE-2116 description, done together with the six sibling PRs. The wire form matches the backend DTOs field for field, the golden shapes and the prediction-time fixtures are identical to the sibling copies, and the `serving_keys` and spine-group deprecations are consistent. Five inline comments below, plus one general comment.

manu-sj:
The PR carries 64 commits, including `f0a89779 tmp` and about forty fork-sync merge commits from `main`. A rebase onto `upstream/main` so only the ticket commits remain would keep the history readable, unless the plan is a squash merge.

python/hsfs/core/feature_view_engine.py
jimdowling: Both parts fixed in `179b08ae1`. You are right about the consequence: the loadtest case only re-reads the materialised versions, so the in-memory path had no cover.

`_get_training_data` sets `spine_anchored` before it creates the metadata, so an in-memory version records its population the same way a materialised one does.

On the decision, I took the first option: `get_training_data`, `get_train_test_split` and `get_train_validation_test_split` take `spine_df` and pass it through. Refusing on read alone would leave a version that can be created and never read, and the three readers already share `_get_training_data` with the creating methods, so it is one argument each rather than a new path. The guide says a spine-anchored version needs the frame again and names these three alongside `recreate_training_dataset`.

Tests: `test_an_in_memory_training_dataset_records_the_spine_too` for the flag, and the existing population checks cover the refusals.

python/hsfs/constructor/prediction_times.py
jimdowling: Both fixed in `179b08ae1`. The class example passes the schedule through `spine_df`:

```python
entities = pd.DataFrame([{"country": "SE", "city": "Stockholm", "street": "Sveavagen"}])
schedule = PredictionTimes.every("daily", offset="08:00", count=7)
fv.get_batch_data(spine_df=schedule.cross(entities, event_time="date"))
```

`_from_user_input` is deleted, along with `TestFromUserInput`, which was its only caller.

python/hsfs/constructor/prediction_times.py
jimdowling: Fixed in `179b08ae1`, both halves.

`_to_local_naive` keeps seconds and microseconds. Its two call sites are better for it: `of()` returns the instant it was given, so the latest-values recipe reads as of now rather than the start of the minute, and a schedule bound of 08:00:30 now excludes the 08:00 slot instead of including a time before `start`.

An integer is epoch milliseconds, the unit `spine_df` reads an integer column in and the unit the feature store keeps event times in. The docstring says so, and `TestPrecisionAndUnits` pins the instant, the millisecond and the schedule bound.

python/hsfs/constructor/inference_spine.py
jimdowling: Fixed here in `179b08ae1`, since the client is the smaller change. `recognized` is now the root features the view selects, plus its primary keys and the event time, which is the set `InferenceSpineResolver` builds from the stored view.

An unselected root column is refused before the frame is typed and uploaded, and with the same accepted-columns list the backend would give. Three tests: a selected root feature is still a passed feature, an unselected one is refused here, and under `allow_passthrough` it is a label like any other column the view does not define.

python/hsfs/feature_view.py
jimdowling: Fixed in `179b08ae1`, and on the nine sibling methods that take `spine_df` rather than only here.

`_reject_unknown_kwargs` allows `primary_keys`, the deprecated spelling these methods keep `**kwargs` for, and refuses the rest with a `TypeError`. `prediction_times` gets the pointer you asked for:

```
get_batch_data() got an unexpected keyword argument 'prediction_times'. `prediction_times` was
replaced by `spine_df`: build the frame with `PredictionTimes.cross(entities, event_time=...)`
and pass it as `spine_df`.
```

The card that emitted it is fixed too, in logicalclocks/hopsworks-front#2103.

jimdowling:
All five inline comments are fixed in `179b08ae1` and answered on their threads.

On the history: merge commits are disabled in this repo, so this lands as a squash and the 64 commits, `tmp` and the fork syncs included, collapse into one whose subject is the PR title. Nothing on `main` will show them. I have left the branch as it is rather than rebasing, because a force-push here has stranded a PR head before and the merged history is the same either way. Say the word if you would rather review it commit by commit and I will rebase onto `upstream/main`.

python/hsfs/constructor/inference_spine.py
```diff
@@ -0,0 +1,474 @@
...
+# The types a root feature group's event time may have for a spine to bind a prediction time to
+# it. The spine's event time column is written in this type, so the two compare as instants. An
+# integer type holds epoch milliseconds, which is the convention every other event-time filter in
+# the feature store uses for it.
+_EVENT_TIME_TYPES = {"timestamp", "date", "bigint", "int", "integer"}
+
+_DECIMAL = re.compile(r"^decimal\((\d+),\s*(\d+)\)$", re.IGNORECASE)
+
+
+def _pyarrow_type(hive_type: str) -> Any:
```
manu-sj: `_pyarrow_type` (line 58) and `_PASSTHROUGH_TYPE` (line 88) are the two legs of one round trip through the offline type vocabulary, and both already exist in `hopsworks_common/core/type_systems.py`, which `hsfs/core/type_systems.py` re-exports:

- `_convert_offline_type_to_pyarrow_type` covers what `_pyarrow_type` does, and also builds `array<...>` and `struct<...>` recursively rather than falling through to `pa.string()`.
- `PYARROW_HOPSWORKS_DTYPE_MAPPING`, through `_convert_pandas_dtype_to_offline_type`, covers the inference direction. It keys on Arrow types instead of `str(dtype)`, so it needs no hand-written entry per spelling, and it enumerates the timezones.

Two consequences are reachable today, both checked against `4fe6014`.

A passthrough column in any timezone other than UTC. `str(dtype)` is `datetime64[ns, Europe/Stockholm]`, which the table does not carry, so `_hive_type` falls back to `string` and `_arrow_table` raises `ArrowTypeError: Expected a string or bytes dtype, got datetime64[ns]`. The caller is told a value does not convert to the type the feature view declares, when what is missing is a table entry. That is the same failure the comment above `_PASSTHROUGH_TYPE` records for the nullable dtypes, arriving by the other door.

A passed root feature of complex type. A view that selects an `array<float>` embedding, supplied in the frame rather than looked up, is written as `pa.string()` here, while the resolver declares the store's own `array<float>` for that column (`InferenceSpineResolver` line 248). The file build fails first, with `Expected bytes, got a 'list' object`.

The millisecond timestamp looks like a real divergence and worth keeping: the shared converter returns `timestamp[us]`, and the description records a defect where a wall-clock timestamp too precise for a millisecond event time made the file disagree with its declaration. Keeping that as a named override on top of the shared converters, rather than as a private copy of both tables, closes the two gaps above and removes more code than it adds.

manu-sj (commented):
One comment, from a second pass over the round-1 fixes. It is about a file this round did not change, so it is a miss from my first review rather than a regression.

manu-sj:
This pull request also reports a conflict, which the description does not mention. It is `python/hsfs/feature_store.py` in `create_feature_view`: upstream made `logging_enabled` and `extra_log_columns` depend on `logging_transport`, and this branch adds `max_feature_age=` to the same call. Keeping upstream's two lines and adding the third resolves it.

As with logicalclocks/hopsworks-ee#3339, this is what stops CI running here, so nothing has reported on the last two rounds. I ran the four changed test files locally against `4fe6014a`: 282 passed. I also ran the Spark golden shapes, which are the only check that this round's quoted struct field names analyse in Spark, since the golden-file comparison only compares text: 20 passed on pyspark 4.1.1.

python/hsfs/constructor/inference_spine.py
jimdowling: Both gaps reproduce, and both tables are gone.

Checked against the current head before changing anything:

```
dtype datetime64[ns, Europe/Stockholm] -> private table: MISSING (falls back to string)
                                          shared mapping: timestamp
array<float>            private=string  shared=list<item: float>
struct<a:int,b:string>  private=string  shared=struct<a: int32, b: string>
```

`_pyarrow_type` now delegates to `type_systems._convert_offline_type_to_pyarrow_type`, and the dtype table is replaced by `_offline_type_of`, which converts the column to an Arrow type and calls `_convert_pandas_dtype_to_offline_type`, so it keys on the Arrow type rather than on a spelling of `str(dtype)`. The millisecond timestamp stays, as the one named override on top, for the reason you identified: the shared converter returns microseconds and the event time is written in milliseconds because a wall-clock value too precise for a millisecond event time made the file disagree with its own declaration.

After the change: the Stockholm dtype maps to `timestamp`, `array<float>` to `list<item: float>`, a nullable `Int64` column to `bigint`, and `timestamp` still to `timestamp[ms]`. `_frame_dtype` and the `re` import went with the tables, so this removes more than it adds.

Client suite 5245 passed, 15 skipped.

manu-sj (approved):

PR #1170, opened 2026-09-15
[HWORKS-3252] hops setup: repoint to a new cluster; add hops logout

MagicLex:
## Problem

`hops setup` resolved the host as `cfg.host or prompt(...)`. Once `~/.hops.toml` held a host, the prompt was suppressed, so a plain re-run (and the dead-key "signing in again" auto-retry) kept hitting the cached host. When that cluster died or moved, there was no way to switch short of `--host` or hand-editing the config. `config.clear()` existed but nothing wired it, so there was no clean reset.

## Change

- `_resolve_host`: on a terminal, prompt for the host with the cached one as the default, so a moved or dead cluster can be repointed without knowing `--host`. Scripts and CI (no TTY) keep the cached host, so non-interactive re-runs never block on a prompt. Switching clusters drops the stale key/project/project_id/feature_store_id (previously `project_id`/`feature_store_id` leaked across clusters).
- `hops logout`: wires the existing `config.clear()` for a clean credential reset. No-op with a clear message when not logged in or in internal (pod) mode.

## Tests

- `_resolve_host`: flag wins, non-interactive reuses cache, interactive prompt switches cluster and drops stale creds.
- `hops setup --force` on a terminal repoints and runs the flow against the new host.
- `hops logout`: clears the profile, no-op when logged out, leaves pod credentials alone in internal mode.

`ruff check` / `ruff format` clean; `pytest python/tests/cli/test_setup.py python/tests/cli/test_logout.py` → 22 passed.


MagicLex:
Why `--force` did not repoint before:

`--force` and host resolution were unrelated. `--force` only skipped one branch, the cached-key short-circuit ("key still works, exit"):

    if not force and cfg.is_authenticated() and _cached_key_works(cfg):
        return None                      # --force only skips this

    host = cfg.host or click.prompt("Hopsworks host", default=config.DEFAULT_HOST)

The host came from that second line, which `--force` never influenced. With a cached host (truthy), the `or` short-circuited and `click.prompt` never ran, so `--force` re-ran the browser token flow against the same cached host. It minted a new key, never a new cluster.

The fix moves host resolution into `_resolve_host`, which on a terminal prompts with the cached host as the default regardless of that short-circuit, so `--force` (and the dead-key auto-retry) can now switch clusters.

jimdowling (approved):

PR #1172, opened 2026-09-16
[HWORKS-3103] Expose the advertised vLLM image tags via REST/SDK and stop re-validating the tag on restart

javierdlrm:
## Summary

The `vllm_image_tag` parameter asked users to supply a tag "made available by a cluster administrator" with no way to find out what those were.

- Add `ModelServing.get_vllm_image_tags(variant="VLLM")`, returning the tags the cluster advertises for that variant, newest first, so the first element is the tag a deployment gets when none is given. The variant argument reuses the existing `PREDICTOR.VLLM_VARIANT_*` constants rather than introducing an enum.
- Document on both `vllm_image_tag` parameters that a deployment keeps its tag even after that tag stops being advertised, so the returned list can be missing the tag an existing deployment runs on.

Requires the matching backend endpoint (`GET /project/{id}/serving/vllmImageTags`, logicalclocks/hopsworks-ee). Against an older cluster the call fails as an unknown path.

## Test plan

- [x] `pytest python/tests/test_model_serving.py python/tests/core/test_serving_api.py` — 94 pass, covering the request path, variant selection, an empty advertised list, and `ValueError` on an unknown variant
- [x] `ruff check` and `docsig` clean on the touched files
- [x] Run against a live cluster from an in-cluster pod with this branch pip-installed from git: ordering, the default tag, the `ValueError`, and a deployment correctly reporting a tag absent from the advertised list
- [x] A notebook exercising the same flow is in the test project on that cluster


robzor92 (commented):
Reviewed alongside logicalclocks/hopsworks-ee#3348 and logicalclocks/hopsworks-front#2111. This one looks clean: `Any` is imported in `serving_api.py`, `PREDICTOR.VLLM_VARIANT_*` exist, the decorator and validation shape match `get_inference_endpoints` and `_validate_deployment_status`, and the apigen alias modules are gitignored so nothing needs regenerating for a new `@public` method.

One doc gap. The docstrings now explain that a deployment keeps a tag after that tag stops being advertised, but not the other half of the new backend contract: `vllm_image_tag=None` on an **update** means "unchanged", not "re-pick the cluster default". In `ServingUtil#validateVLLMUserInput` a null tag on the update path inherits the stored value and returns before the advertised-list check ever runs, so after this change there is no way for an SDK user to say "move me to the current default" except naming a tag explicitly. Worth a sentence on the `vllm_image_tag` parameters, since the current wording ("`None` uses the cluster default, the newest advertised tag") reads as if it still re-defaults on save.

For transparency: I read this against the backend rather than running the Python suite, so the 94-pass claim in the test plan is untested by me.

javierdlrm:
Doc gap confirmed and fixed in f2044507e. Your reading of `validateVLLMUserInput` is right: on the update path a null tag inherits the stored value and returns before the advertised-list check runs, so "move me to the current default" has no SDK spelling other than naming the tag.

Rewrote the `vllm_image_tag` parameter docs on both `Model.deploy` and `ModelServing.create_predictor` to separate the two cases: on a new deployment `None` selects the newest advertised tag of the variant, on an update `None` means "unchanged" and moving to the cluster default takes naming that tag explicitly. Kept the sentence about a deployment keeping a de-advertised tag, since that is the other half of the same contract.

Folded in Copilot's point on the same lines while I was there: the text now says the tag must come from the list returned for the matching `vllm_variant`, and notes that an unqualified `get_vllm_image_tags()` lists the standard vLLM tags. Also split the one-line docstring across lines. The backend `@Schema` descriptions for both fields got the matching treatment in logicalclocks/hopsworks-ee#3348, so the generated REST docs and the SDK docs now say the same thing.

The asymmetry is deliberate rather than an oversight: the wire cannot tell an omitted field from an explicit null, so null cannot mean both "re-pick the default" and "leave it alone", and silently moving a running deployment to a different runtime image during an unrelated edit is the worse of the two failures. A first-class "follow the default" would want its own signal.

Fair call on the test plan. Re-ran here: 94 passed (`test_model_serving.py` and `core/test_serving_api.py`), ruff check and format clean, docsig reports only the three pre-existing SIG305s in `model_serving.py` at lines 241, 361 and 791, none of them on the lines this PR touches.

Separate commit for re-review; will squash before merge.

robzor92 (approved):
Approving. The doc gap is closed on both `Model.deploy` and `ModelServing.create_predictor`: the parameter docs now separate create-time defaulting from update-time inheritance, say that moving to the cluster default takes naming the tag explicitly, and keep the de-advertised sentence. Copilot's variant-ambiguity nit is folded into the same lines, so the text now points at the list for the matching `vllm_variant` and notes what an unqualified `get_vllm_image_tags()` returns. Matches the `@Schema` wording in logicalclocks/hopsworks-ee#3348, so the REST and SDK docs agree.

One CI note so it does not read as a real failure: `Unit Tests (Windows) (Local TZ)` is red with 1 failed / 4559 passed, and the failure is `SparkException: Python worker failed to connect back` / `SocketTimeoutException` — a Spark-on-Windows runner flake on a path this PR does not touch. The branch was rebased after that run, so it re-runs anyway.

Squash before merge as planned.

PR #1173, opened 2026-09-16
[HWORKS-2879][APPEND] Register Hudi's Spark session extension in Spark Connect sessions

robzor92:
Follow-up to the Hudi 1.2.0 / Spark 4.1 upgrade. Companion to hopsworks-helm#2364, which fixes the same defect cluster-side.

## What is broken

The SDK's Spark Connect session builders set `spark.sql.extensions` to Delta's extension only, so `HoodieAnalysis` is never injected into the session. Its rule is what rewrites an incremental scan to carry a filter on `_hoodie_commit_time`; without it Spark prunes that meta column away, and Hudi's file-group reader then applies the incremental filters against a schema that no longer contains it — dropping every record.

The result is that `as_of(exclude_until=)` and `read_changes()` on a HUDI feature group return **no rows**, and `count()` returns 0 because it prunes to zero columns. Measured on a 5.1 cluster, one incremental DataFrame over 3 commits returned 3 rows from `collect()` but 0 from both `count()` and `select("id")`; with the extension registered, a one-commit window returns exactly 1.

Three call sites, all guarded by `_is_spark_connect_env()` plus the `spark-delta` engine branch:

- `python/hopsworks/spark.py` — `build_spark()`
- `python/hopsworks_common/client/external.py` — `spark-no-metastore` and `spark-delta` engines
- `python/hsfs/engine/spark_no_metastore.py`

## Why this is safe where the jar is missing

`build_spark()` documents the terminal-spark image as the supported environment, and that image carries `hudi-spark4.1-bundle_2.13-1.2.0.jar` on the Spark classpath alongside `delta-spark_2.13-4.3.1.jar`.

For the other documented path — a bare `pip install pyspark[connect]` venv, where no Hudi jar exists — nothing breaks: `SparkSession.applyExtensions` catches `ClassNotFoundException` and `NoClassDefFoundError` and only logs a warning. Verified directly in the shipped `spark-sql_2.13-4.1.3.0` bytecode rather than assumed.

## Scope

`hsfs/storage_connector.py` and `hsfs/core/delta_engine.py` also mention this property, but only to *read* it for user-facing warnings, so they are untouched.

`branch-5.1` (Hudi 1.2.0) and `branch-5.0` (Hudi 1.0.2.2) carry the same Delta-only builders and are affected too — backports to follow separately if wanted.

https://claude.ai/code/session_016XaT56wVgdRwKACjphgehk


robzor92:
### When this actually takes effect

`spark.sql.extensions` is a **static** config: against an already-running Connect server the client's attempt is refused (`CANNOT_MODIFY_STATIC_CONFIG` — the warning is documented in `skills/data/hops-spark/SKILL.md`). So this change governs only sessions the builder brings up itself (locally spawned Connect server, in-process session in the terminal-spark image), exactly like the Delta entry beside it. The remote case is fixed server-side by hopsworks-helm#2364, whose config the Connect entrypoint copies before starting the server. This PR is the consistent-but-secondary half.

**Compatibility checks:** `delta_engine.py` and `storage_connector.py` both test for the Delta extension by substring, so the comma-joined list still satisfies them; no test asserts the renamed `_DELTA_EXTENSIONS`; and a missing Hudi jar cannot break session creation — `SparkSession.applyExtensions` catches `ClassNotFoundException`/`NoClassDefFoundError` and only warns, verified in the shipped `spark-sql_2.13-4.1.3.0` bytecode.

**Scope:** the requirement is new in Hudi 1.2.0 — isolated on a fixed Spark 3.5, upstream 1.0.2 works without the extension and upstream 1.2.0 does not — so `branch-5.0` (Hudi 1.0.2.2) needs no backport.

manu-sj (approved):

PR #1164, opened 2026-09-13
[HWORKS-3240] Feature view inference logging: real-time online logs from the Arrow inference logger

jimdowling:
## Summary

Feature logging v2 client: one transport per feature view (`realtime` through the inference logger, `job` through a file buffer on the pod, the Dataset API and a packaged commit job), `DeploymentLoggingConfig` overrides, `read_log(online=True)`, the hour/day materialization interval, and Prometheus metrics published by the predictor for the job transport so the deployment page card serves both transports.

## Changes

- `hsfs`: `FeatureLogging.transport`, `enable_logging(transport=)` refusing the other transport, `delete_log(transport=)` to switch, `read_log(online=True)`, `create_feature_view(logging_transport=)`, `feature_log_commit_job.py` (manifest-first claims, one Delta application transaction per chunk so a chunk staged twice under any claim commits once, a part the group's schema rejects parked under `failed/`, stale upload adoption, self re-trigger), Delta write commit properties carrying one transaction or a list; chunks keep their Arrow types through the commit, including `map<>` features, so nested integers, decimals and all-null dates land as declared.
- `hsml`: `hsml.deployment` package with shims on the released paths, `DeploymentLoggingConfig`, default predictor batching to the receiver's row and byte limits, job transport writer subprocess (`hopsworks_common.core.feature_logging_file`) with a byte-bounded frame queue and a single shutdown budget, `_FileLogCollector` metrics, `commit_feature_logs()`.
- `utils/python/hsfs_utils.py`: idempotent append for the realtime materialization (Delta application transaction, pending offset range).
- Codex review (13 findings) addressed; see the squashed commit body.
- `hsml`: the API protocol is the caller's choice and still defaults to REST, and a model deployment served by the default predictor now serves gRPC too when `api_protocol="GRPC"` is passed. Rows then travel as one KServe v2 tensor per schema field, typed where the column's values allow it and JSON text in a `BYTES` tensor where they do not, so nulls, strings, timestamps, nested values and integers past 2**53 survive; a field no row of the batch carries is left out rather than sent as a column of nulls, and a batch whose rows differ in the optional fields they carry travels as one tensor of whole rows. Response tensors take their datatype from the values, so an integer prediction comes back an integer. `deployment.predict()` validates the rows against the schema before encoding and decodes the response tensors back into the dictionary a REST call returns, so calling code does not change with the protocol. The gRPC channel is opened once under a lock instead of once per racing caller.

## Tests

[...]

PR #1198, opened 2026-09-21
[HWORKS-3240][Append] Keep the pending offsets file out of the feature group's Delta location

jimdowling:
https://hopsworks.atlassian.net/browse/HWORKS-3240

One of the four regressions behind the 134 failures in the 2026-09-20 helm nightly (run 35527900893). Sibling PRs: docker-images (transformer 501 with several workers), loadtest (`Feature` attribute access), hopsworks-ee (image pipeline pinned to one commit).

## Problem

`*_offline_fg_materialization` failed permanently for realtime feature-logging groups (32 failures): `read_log()`, `get_log_timeline()` and the Arrow Flight reads then saw `No delta logs found for featuregroup ..._log_1`. #1164 writes the offsets a run is about to consume to `<location>/kafka_offsets_pending` before `entity.insert()`. A logging group is liquid-clustered, so its first run reached `DeltaTable.createIfNotExists().location(loc)` with a directory that already held a file and no `_delta_log`, which Delta refuses with `DELTA_MISSING_DELTA_TABLE`. The pending file stayed behind, so every later run failed the same way. Upserts were unaffected: their `kafka_offsets` write happens after the insert.

## Fix

Write the pending range beside the feature group directory, `<location>_kafka_offsets_pending`, instead of inside it. The fence keeps its semantics (written before the Kafka read, removed after the offsets file is saved).

## Review follow-up (Codex round 1)

Two findings, both fixed in the second commit:

- **P1, no migration for existing claims.** A cluster that already hit the bug has `<location>/kafka_offsets_pending` in a table directory Delta never created, and an append interrupted under the old client leaves a claim there that the new code did not read. `_pending_offsets()` now reconciles the old place first: removed when there is no `_delta_log` (nothing was appended, the file only blocks the create) or when a newer claim exists; moved to the new place otherwise, so the retry repeats that range and Delta dedups on its transaction version.
- **P2, the claim outlives the group.** The file sits beside the table directory, which is all a delete removes, so a group recreated under the same name and version inherited the old range. The file name now carries the feature group id (`<location>_kafka_offsets_pending_<id>`); a recreated group has a new id and never reads the old claim. The old claim is left behind as a small orphan, since the backend deletes only the table directory. (A topic-name check was considered and rejected: `Utils.getFeaturegroupTopicName` falls back to a project-wide topic, so two groups can share one.)

Exercised locally with pyspark 4.0.1 + delta-spark 4.0.0: legacy claim in a never-created table dir, legacy claim on an existing table, legacy and new claim together, a deleted group's claim beside a recreated group, a plain new claim, and no claim; the clustered `createIfNotExists` succeeds in all six and the returned range is the expected one. Not re-run on the cluster: the full-suite Job there is using the first commit's file through the `fs_py_job_util` override and swapping it mid-run would change what that run tests.

## Verification

Reproduced locally with pyspark 4.0.1 and delta-spark 4.0.0: a clustered `createIfNotExists` at a location holding `kafka_offsets_pending/` fails with `[DELTA_MISSING_DELTA_TABLE]`; the same table with the file beside the location is created and appended to.

Loadtests: run on jim-bugs-sep21 (chart 5.2.0-alpha-1, images 5.2.0-SNAPSHOT) as in-cluster Jobs. `hsfs_utils.py` is baked into the spark image, so the cluster ran this branch's file through `fs_py_job_util=hdfs:///user/spark/hsfs_utils_fix.py` plus a variables refresh.

- [x] `workflows/model_serving/python_driver/external/test_feature_logging_transport.py` passed: the realtime `iris_schema_1_log_1` group, liquid-clustered, got its `_delta_log` on the first materialization and its `read_log()` returned rows.
- [x] `workflows/model_serving/python_driver/external/test_feature_logging.py`: 2 passed, 5 failed. All 7 `fv_logging_test_online_1_log_1_offline_fg_materialization` runs succeeded (the nightly failed 12 of 12 here). The Delta history of that group is one `CREATE TABLE` plus one 10-row `WRITE` per run with distinct `txn` versions (70, 100, 130, 160, 190), so the fence still dedups and nothing was appended twice; no `_kafka_offsets_pending` is left behind. The 5 failures are downstream of materialization: `test_implicit_logging` asserts `label_encoder_fg1_feature3_` value 1 where it expects 2 (a transformed value computed by the predictor, not by this job), and the four tests after it then read that test's leftover rows (20 and 30 rows where 10 are expected) because the failed test never reached its `delete_log()`. Rerun alone in a fresh project, `test_implicit_logging` passes (1 passed in 798s, two materializations). The failure is order-dependent and predates this branch: every test in the module draws new random rows and makes a new training dataset, but the predictor scripts pin `init_serving(training_dataset_version=1)`, so from the second test on the deployment encodes labels with the first test's statistics while the test expects the current dataset's; a category missing from one of the two shifts the codes by one. The nightly never reached these assertions because the materialization failed first. Left for a loadtest follow-up.

Not covered: `test_deployment_schema.py` (the `iris_schema` materializations of the nightly beyond the transport test) and the feature_store `test_feature_logging.py` job-transport tests, whose groups are not clustered and never hit this path.


jimdowling:
## Full loadtest suite on jim-bugs-sep21 (in progress)

`workflows/` as an in-cluster Job (`dev_scripts/deploy_loadtest.sh jim-bugs-sep21 --workers 3 workflows`, `CONNECTORS` set for `SQL_CLICKHOUSE`), 1039 tests collected against the nightly's 1052. The cluster runs chart 5.2.0-alpha-1 with images `5.2.0-SNAPSHOT`, plus the four fixes under test: the `loadtest` branch of #1060 in the Job, `hsfs_utils.py` from hopsworks-api#1198 through the `fs_py_job_util` variable, and the KServe wrapper of docker-images#1065 overlaid onto the `pandas-inference-pipeline` and `minimal-inference-pipeline` image tags in the in-cluster registry. The `hopsworks` EAR is the one nexus republished on 2026-09-21 08:43 UTC, which is past ee `6e41afb67`, so hopsworks-ee#3375's mismatch is not present on this cluster either way.

**Status at 446 of 1039: 325 passed, 99 skipped, 22 failed.** None of the 22 is one of the four nightly failure signatures (`'Feature' object is not subscriptable`, `501 Could not find PredictorConfig`, `PUT /api/tags/{name}/archive` 404, `No delta logs found for featuregroup ..._log_1`). The 99 skips are the data source connector types with no `CONNECTORS` entry on this cluster.

Failures so far, tracebacks pending (the in-cluster runner prints them at the end):

- `python_driver/test_feature_group_transformation_functions.py`: 17 on-demand transformation cases (`test_fg_no_on_demand_transformation`, `test_one_to_one_*`, `test_many_to_one_*`, `test_many_to_many_*`, `test_multiple_on_demand_*`, `test_odt_chaining_drop_intermediate`, `test_on_demand_transformation_with_model_dependent_transformations*`)
- `pyspark_driver/test_feature_group_transformation_functions.py::test_case[test_streaming_feature_group_on_demand_transformations]`
- `python_driver/test_feature_group_insert.py::test_multi_part_insert[True-DELTA]`
- `python_driver/test_sql_transaction_limits.py::test_rate_per_sec[True]`
- `pyspark_driver/test_external_feature_group.py::test_pyspark_external_feature_group`
- `pyspark_driver/test_feature_logging.py::test_pyspark_feature_logging`

The cluster is saturated while these run: 15 Spark applications (10 running, 5 submitted) and 17 Pending pods across the two worker nodes, so some of the above may be timeouts under load rather than code; they are triaged against the nightly's baseline once the tracebacks are in. This comment is updated when the Job finishes.

The targeted runs that verify this PR's own change are in the PR description.

robzor92 (approved):

PR #1199, opened 2026-09-21
[FSTORE-2107] missing init kafka offsets

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```markdown
```


dhananjay-mk (commented):
There is one data-loss race in a first materialization run with neither a caller-supplied nor a saved checkpoint. If a partition has no record at or after the creation-time floor when the timestamp lookup runs, a concurrent producer can append a record before the later end-offset lookup. Using that newer end offset as the starting offset skips the record. This is limited to the no-checkpoint path; a Python SDK insert that launches the job with an explicit initial checkpoint does not use this fallback. Please capture a conservative end offset before the timestamp lookup (or otherwise ensure the fallback cannot pass a concurrent write), and cover this interleaving in both implementations.

python/hsfs/core/kafka_engine.py
```diff
@@ -263,6 +263,70 @@ def _kafka_get_offsets(
...
+            if result.error is not None:
+                # Reading from the low watermark is what this function is meant to avoid,
+                # but a partition Kafka would not answer for has no offset to trust, and
+                # re-reading records is recoverable where skipping them is not.
+                offset = low
+            elif result.offset < 0:
+                # Kafka answers a timestamp past the last record with a negative offset:
+                # every record the partition holds is older than the timestamp, so the
+                # reader belongs at the end of it.
+                offset = high
```
dhananjay-mk, line 319: When `offsets_for_times` returns a negative offset, `high` is fetched afterward. If the lookup sees no matching record at end offset 10, then this feature group's producer writes record 10 before `get_watermark_offsets`, `high` becomes 11 and starting at 11 skips that row. This can happen on a first manual or scheduled materialization with no checkpoint. Use an end offset captured before the timestamp lookup, or another conservative start.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/engine/hudi/HudiEngine.java
```diff
@@ -324,10 +330,73 @@ private String getCheckpointTopic(HoodieTimeline commitTimeline) {
...
+    }
+    Map<org.apache.kafka.common.TopicPartition, OffsetAndTimestamp> found =
+        consumer.offsetsForTimes(lookups);
+    Map<org.apache.kafka.common.TopicPartition, Long> endOffsets = consumer.endOffsets(topicPartitions);
+
+    Map<org.apache.kafka.common.TopicPartition, Long> startOffsets = new HashMap<>();
+    for (org.apache.kafka.common.TopicPartition tp : topicPartitions) {
+      OffsetAndTimestamp atOrAfter = found.get(tp);
+      if (atOrAfter == null) {
+        startOffsets.put(tp, endOffsets.get(tp));
```
dhananjay-mk, line 383: The Hudi path has the same narrow race: `offsetsForTimes` can return null for a partition, then a matching record can arrive before `endOffsets` runs. The new end offset is past that record, so this initial checkpoint skips it. Capture a conservative end offset before the timestamp lookup, and test an append between the two calls.

